PR #664, opened 2026-07-30
fix(ui): repair the dead agent link and humanize wire enums for visitors

### LINKED ISSUE #641: Broken link on the public Run stats page, plus raw enum values shown to visitors
### What is wrong

Warren runs a **public read-only dashboard**, so these are visible to anonymous visitors. Two related problems on the same pages.

**1. A broken link (silent failure).** In `src/ui/src/pages/run-analytics/Tables.tsx` (around lines 29-37), agent rows link to `/agents/${name}`. No such route is registered in `App.tsx` — only `/agents`. The catch-all route bounces the visitor to `/runs` with no explanation, so clicking an agent name appears to do something random.

Fix by linking to `/agents`, or by adding the detail route if you think that is the better answer. Say which you chose and why.

**2. Raw enum values and internal jargon.** These reach visitors unformatted:

- `src/ui/src/components/StatusIndicator.tsx` (~line 81) renders the raw enum `pr_open` as a badge label. Should read "PR open".
- `src/ui/src/pages/RunDetail.tsx` (~lines 126-129) and `src/ui/src/pages/run-analytics/Charts.tsx` (~lines 213-233) render failure reasons raw: `finalize_failed`, `spawn_failed`, `evicted`. These need a label map — keep the raw value in a tooltip so operators can still see it.
- `src/ui/src/pages/PlanRuns.tsx` (~lines 338-354) renders a child summary as `2p / 0i / 3m / 1f`. Undocumented single-letter jargon in a public table. Either spell it out ("2 pending, 3 merged, 1 failed") or add a tooltip.

### What to change

Fix the link, then add human-readable labels for the enum values. Keep the mapping in one place rather than scattering string literals across components.

### Done when

Every agent row links somewhere real, and no raw underscore-separated enum appears in visitor-facing text.

### Good to know

You can see the live public instance at <https://app.warren.run> to view these pages as an anonymous visitor. Line numbers are approximate — search for the strings. A nice self-contained React/TypeScript task with visible results.

---

**New here?** Warren is a self-hosted control plane for coding agents (Bun + TypeScript). Setup and conventions live in [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md).

Before opening a PR, run the full gate suite:

```bash
bun run check:all   # or: bun run verify
```

It runs 12 gates (lint, typecheck, dedup, deps, file size, debt markers, bundle size, generated-doc checks, coverage, CI parity). `CHECK_ALL_VERBOSE=1` streams full output; `--bail` stops at the first failure.

Tracked internally as `warren-14fc`. Questions are welcome on this issue — a maintainer will help you scope it.

EZArchy:
## Summary

Closes #641 (tracked internally as `warren-14fc`). Two related problems on the
public read-only dashboard, where everything below is visible to anonymous
visitors.

- **Fixes a dead link.** `run-analytics/Tables.tsx` linked agent rows to
  `/agents/:name`, a route `App.tsx` never registers — the catch-all silently
  bounced the visitor to `/runs`.
- **Adds one home for visitor-facing labels.** New `src/ui/src/lib/labels.ts`
  turns wire enums into prose. The raw value always survives in a `title`
  tooltip so an operator can still read what the server recorded, and unknown
  values degrade through `humanizeWireValue` instead of leaking underscores.
- **Chose `/agents` over a new detail route.** There is no agent-detail page;
  the Agents list already renders the full agent envelope in an expandable row,
  so a detail route would be a second surface for the same data and a much
  larger change than this bug warrants. To keep the link meaningful rather than
  dumping every agent on the same undifferentiated list, rows link to
  `/agents?agent=<name>` and `Agents.tsx` pre-expands that row at mount.

| Surface | Before | After |
|---|---|---|
| Run stats, agent row | `/agents/claude-code` → bounced to `/runs` | `/agents?agent=claude-code`, row pre-expanded |
| Plan-run child badge | `pr_open` | `PR open` |
| Run detail header | `finalize_failed` | `Finalize failed` |
| Failure-reason pie (legend, slice labels, tooltip) | `oom_killed`, `evicted` | `Out of memory`, `Evicted` |
| Plan-run detail header | `pr_closed_without_merge` | `PR closed without merge` |
| Plan-run child table | `child_seed_not_found:warren-a` | `Child seed not found (warren-a)` |
| Plan runs, "Children" column | `1p / 2i / 2m / 1f` | `1 pending · 2 in flight · 1 merged · 1 skipped · 1 failed` |

Two judgement calls worth flagging — happy to change either:

- `burrow_run_lost` / `burrow_unreachable` read **"Sandbox run lost"** /
  **"Sandbox unreachable"**. Burrow is the LocalProvider's substrate — an
  implementation detail a visitor has no name for, and one that does not exist
  at all under `WARREN_RUNTIME=k8s`. Raw value still in the tooltip.
- The child rollup gives `skipped` its **own bucket** instead of folding it into
  `merged` the way the letter form did. Once the counts are words, calling a
  child that was never dispatched "merged" is simply wrong.

## Changes

**New**

- `src/ui/src/lib/labels.ts` — `humanizeWireValue`, `formatRunFailureReason`,
  `formatPlanRunFailureReason`, `formatChildStateCounts`. Status-*pill* labels
  are deliberately **not** duplicated here: `components/StatusIndicator.tsx` is
  already the single registry for `{label, variant, icon, pulse}`.
  Plan-run failure text is *not* an enum — the coordinator writes free-form
  `reason` or `reason:detail` — so there is no closed set to map exhaustively;
  that formatter humanizes the discriminator half and keeps the detail verbatim
  in parentheses.
- `src/ui/src/lib/labels.test.ts` — 14 cases. The exhaustiveness case walks the
  canonical `RUN_FAILURE_REASONS` tuple from `src/core/wire.ts` (not a local
  copy) and asserts every reason has a prose label containing no underscore, so
  a reason added to the kernel without a label fails the suite. The map is also
  `satisfies Record<RunFailureReason, string>`, so the same gap fails
  `typecheck`.

**Modified**

- `pages/run-analytics/Tables.tsx` — the link fix.
- `pages/Agents.tsx` — reads `?agent=<name>` to pre-expand a row.
- `components/StatusIndicator.tsx` — `pr_open` → `"PR open"`; the unknown-status
  fallback humanizes; new `<RunFailureBadge>`.
- `pages/RunDetail.tsx`, `pages/run-analytics/Charts.tsx`,
  `pages/PlanRunDetail.tsx`, `pages/plan-run-detail/child-table.tsx`,
  `pages/PlanRuns.tsx` — consume the formatters.
- `scripts/file-size-budgets.json` — `RunDetail.tsx` was at its frozen ceiling,
  so the failure badge moved into `StatusIndicator.tsx` next to the other
  wire-value badges rather than growing the page. Ratchets **down**, 1103 → 1101.
- `scripts/bundle-size-budgets.json` — re-baselined via
  `bun run check:bundle-size --update` on a clean `--frozen-lockfile` build:
  +607 B gzip js / +1211 B raw js, well inside `AUTO_RAISE_CAP`; css ratchets
  down. Rationale recorded in a `$comment`, per the file's convention.

## Test plan

- [x] `biome check .` passes
- [x] `tsc --noEmit` passes
- [x] `bun test` — 3572 pass / 21 skip / 4 fail. The 4 are
      `scripts/cloud-armor.test.ts` and fail **identically on a clean `main`** in
      my environment: the test pins `PATH` to `/usr/bin:/bin` and this is a NixOS
      box with no `bash` there. `bash deploy/gcp/cloud-armor.sh --dry-run` emits
      its 8 `+ gcloud` lines normally, so this is environmental and CI's ubuntu
      runner is unaffected. Nothing in this diff touches that area.
- [x] `bun run check:all` — 11/12 gates pass; `check:coverage` reports coverage
      above both floors (functions 89.71% vs 89.30, lines 92.17% vs 91.85) and
      exits non-zero only on the same 4 tests above.
- [x] Manual verification — every surface in the table above was driven in a
      real browser against a local warren with a seeded database, both as an
      operator and anonymously. The agent link was verified by clicking it:
      it lands on `#/agents?agent=claude-code` with that row expanded.

---

**Separate pre-existing bug, found while verifying this one (not fixed here).**
The Run stats page **crashes for anonymous visitors** on `main`, including on
<https://app.warren.run>. The public projection drops `costUsd` / `priced` from
each `RunGroupBucket` and `cost` from `totals` (`REDACTED_RUN_GROUP_FIELDS` /
`REDACTED_RUN_TOTALS_FIELDS`), but `src/ui/src/api/client.ts` types all three as
required, so `KpiCards` throws `Cannot read properties of undefined (reading
'total')` and `Tables.tsx` throws `… (reading 'toFixed')`. The whole page falls
through to the error boundary. Verified against the live instance:
`curl https://app.warren.run/analytics/runs?days=7 | jq .byAgent[0]` has no
`costUsd`. It is a wire-type drift rather than a labelling problem, so I left it
out of this PR — say the word and I'll open an issue or a follow-up PR.

### DIFF
--- a/scripts/bundle-size-budgets.json
+++ b/scripts/bundle-size-budgets.json
@@ -6,20 +6,21 @@
 	"$comment_pl55a3_closeout_2026_05_29": "warren-bfc6 / pl-369d: pl-55a3 (warren-6358, frontend design-system revamp, v0.7.2) is now closed. The TEMPORARY overhaul ceilings raised across PR #197/#201 are collapsed to honest floors here. NOTE: pl-55a3 grew the bundle right up to its ceilings — the only metric with genuine room to ratchet down is gzip css (8500 → 8200). raw js, raw css, and gzip js are at their true floors and cannot drop. The first attempt at this closeout (commit c2a3ba5) set gzip js to 172580 from a number that was NOT produced by `bun run check:bundle-size` — it matched Vite's build-log reporter (~2KB cooler), so when CI ran the actual script the identical bytes gzipped to 173931 and the guard failed. Re-baselined from the actual check script on a clean --frozen-lockfile build (byte-identical to CI): gzip js 173931 / raw js 583998 / gzip css 8005 / raw css 40432 (largest chunk == totals — single js + single css asset). Budgets = those figures + a small churn headroom. The bulk of the js floor is the Phase 5 motion layer (framer-motion v12 via strict LazyMotion + domAnimation, NOT the full motion proxy). The 'ratchet only goes down' rule is back in force from here — future UI work must hold flat or shrink against THIS floor.",
 	"$comment_plot_ui_deletion_2026_07_26": "RATCHET DOWN (warren-1f12 / pl-3a79 step 10): the plot UI deletion removed ~3.8k LOC of pages and components (plot-detail/, PlotSummary, the Workspace surface, NewPlotButton, PlotStatusBadge, PlotMetaCardContent, DefaultLanding) plus the plotsApi client block and the plot type region. Re-baselined via `bun run check:bundle-size --update` on a clean --frozen-lockfile build: gzip js totals 299685 -> 285902, largest chunk 174236 -> 160480, gzip css 8362 -> 7897. Lowering, so it applies unconditionally. This is the new floor.",
 	"$comment_capability_layer_2026_07_27": "BOUNDED AUTO-RAISE (warren-f53e / pl-b82d step 19): the UI capability layer adds the useCapabilities hook, the OperatorOnly / OperatorRoute wrappers, and ~25 wrap sites across the pages, against which the deleted PLOT_STATUS registry and the RefreshProjectsCTA label prop return only a few hundred bytes. Net +673 B gzip js / +1984 B raw js — ordinary feature growth, well inside AUTO_RAISE_CAP, re-baselined via `bun run check:bundle-size --update` on a clean --frozen-lockfile build. No new dependency. The 'ratchet only goes down' rule resumes from here.",
+	"$comment_public_labels_2026_07_30": "BOUNDED AUTO-RAISE (warren-14fc / #641): src/ui/src/lib/labels.ts adds the visitor-facing label maps for run / plan-run failure reasons plus the plan-run child rollup, and the call sites that consume them. Net +607 B gzip js / +1211 B raw js — ordinary feature growth, well inside AUTO_RAISE_CAP, re-baselined via `bun run check:bundle-size --update` on a clean --frozen-lockfile build. No new dependency. CSS shrank (no new utilities) and is ratcheted down to its measured floor. The 'ratchet only goes down' rule resumes from here.",
 	"totals": {
 		"raw": {
-			"js": 944073,
-			"css": 38509
+			"js": 945284,
+			"css": 37779
 		},
 		"gzip": {
-			"js": 285757,
-			"css": 7899
+			"js": 286364,
+			"css": 7781
 		}
 	},
 	"largest": {
 		"gzip": {
-			"js": 160340,
-			"css": 7899
+			"js": 160918,
+			"css": 7781
 		}
 	}
 }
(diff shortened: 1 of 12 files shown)


jayminwest:
@EZArchy your two empty-state PRs (#661, #664) improved the visitor experience. I filed #806 for the spectator dead ends that those PRs did not cover:

- The Agents page shows operator instructions to anonymous visitors
- The Login page has no path back to the run list
- A paused run shows no explanation of its state.

The work continues your PRs, so I offer it to you first. I plan to merge a set of community PRs early next week.

PR #772, opened 2026-08-03
fix: route-param path traversal, salvage-path disclosure, and the salvage callback gap- #4

gaby:
## Summary

Findings from an audit of the codebase, plus the two review rounds on them. Four fixes, two findings recorded rather than fixed.

**Fixed — path traversal to arbitrary file write (High).** `matchRoute` percent-decodes every `:param`, but the `([^/]+)` capture constrains only the *raw* pathname, and `URL.pathname` preserves `%2F`. So `POST /runs/..%2F..%2Fetc%2Fpwn/salvage` matched the single-segment pattern and decoded to `id = "../../etc/pwn"`, which the salvage intake joined straight onto the salvage directory — up to 32 MiB of caller-supplied bytes written to an arbitrary path, for any bearer holding `dispatch` (short of `admin`). Confirmed end to end, not by inspection: against the unpatched server the new handler test writes outside its `mkdtemp` root.

**Fixed — unhandled `URIError` crashes the request pipeline (Medium).** `/runs/%ZZ/cancel` threw out of `matchRoute`, which `handleRequest` calls *outside* its try/catch. The connection died with ECONNRESET rather than returning warren's error envelope.

**Fixed — salvage events leak the host bundle path to spectators (Medium).** `salvagePath` is a `REDACTED_RUN_FIELDS` member, but `reap.workspace_salvaged` and `reap.workspace_salvage_recorded` served the identical absolute path to anonymous readers as `bundlePath`, and `reap.workspace_salvage_failed` carried it again inside raw git stderr. Third instance of the `warren-5f59` / `warren-d8f4` pattern; scenario 39 had no assertion for it.

**Fixed — in-pod salvage was 403'd by its own auth gate.** `/runs/:id/salvage` was missing from `RUN_CALLBACK_ROUTE_PATTERNS`, so the run-scoped token every pod carries (`warren-57fd`) was refused on the one route that banks a run's otherwise-unrecoverable commits before the `emptyDir` dies. The route's own doc comment already asserted the pod authenticates that way — only the allowlist was stale.

**Recorded, not fixed — path-mode previews share warren's browser origin (`warren-3f8a`, High).** `WARREN_PREVIEW_MODE=path` is the default and serves previews at `/p/{run-id}/`, the same origin the UI keeps the operator token on in `localStorage`. The preview app is code an agent wrote, so a script in it can read that token. No response header closes this: the `sandbox` CSP directive that would give the preview an opaque origin also makes the browser treat its own sub-resource requests as cross-site, which blocks the `SameSite=Lax` preview cookie those requests need. Shipping it would break every path-mode preview. Documented in `SECURITY.md` with subdomain mode as the operational mitigation.

**Recorded, not fixed — the `no_intent` salvage can still be refused (`warren-9d24`, Medium).** Surfaced in review. `pollForIntent` waits 50 min by default while the heartbeat watchdog force-fails a silent run at 45 min, so the terminal `no_intent` POST can land after the row is terminal and 401. The `warren-5ea1` early salvage has already banked a bundle by then, so the loss is the final capture rather than the work. The `push_failed` window is unaffected (it POSTs before `finalize-result`, while reap still awaits), and the terminal-reconcile net never races a live pod. Fixing it means either loosening the liveness gate or coupling two independently tunable timeout knobs — a maintainer's call, not something to slip into this branch.

## Changes

- `src/server/router.ts` — `decodeParam` refuses a param that decodes to a path separator or NUL, and refuses a malformed escape instead of throwing. `matchRoute` and `pathExists` now share one `matchPattern`, so a refused param reports 404 (no such resource) rather than 405.
- `src/runtime/salvage.ts` — new `salvageBundlePath` validates the run id and asserts containment under the salvage dir. Both salvage writers (`src/server/handlers/runs/salvage.ts`, `src/runs/reap/salvage.ts`) route through it, so a future non-route caller inherits the check instead of re-deriving it.
- `src/server/handlers/runs/event-projection.ts` — `bundlePath` / `salvagePath` censor on the key; `reap.workspace_salvage_failed` joins `RAW_FAILURE_EVENT_KINDS`, with the sanitizer redacting `errors` element-wise so the *count* of distinct failures survives. `rescueRef` stays in the clear — it holds only the already-public run id, which is why `PUBLIC_RUN_FIELDS` admits it as `salvageRef`.
- `src/runs/spawn/run-token.ts` — `/runs/:id/salvage` added to the callback allowlist, still pinned to its own run id and still subject to the liveness check.
- `src/server/auth.ts`, `src/runs/spawn/callback-env.ts` — refreshed the two security rationales that still said a run token reaches "three" callback routes.
- `SECURITY.md`, `CHANGELOG.md`, `.seeds/issues.jsonl` — the V1-limitation entry, the release notes, and three tracker rows (`warren-7c1e` closed, `warren-3f8a` and `warren-9d24` open).

Tests: every new assertion was checked against the pre-fix code and fails there — the traversal e2e, the malformed-escape e2e, and all three projection assertions.

## Test plan

- [x] `biome check .` passes (via `bun run lint`, including the four repo guards)
- [x] `tsc --noEmit` passes
- [x] `bun test` passes — 3785 tests, 0 failures
- [x] Manual verification — reproduced the traversal against the unpatched server (wrote `/tmp/pwned.bundle` from a request whose salvage dir was a `mkdtemp` under `/tmp`), and confirmed `bun run check:all` reports 12/12 gates on the rebased branch


gaby:
@jayminwest These were all done using Claude Opus 5 (max effort).

jayminwest:
Awesome work @gaby, thanks for the PR!

jayminwest:
@gaby your PR #772 fixed real security problems. I filed #807 in the same area. Two defaults are wrong for a public instance:

- The server disables the idle timeout on all routes to protect the long-lived NDJSON tail
- The public deployment runs event streams with no lifetime limit.

The fix is small and scoped. Tell me if you want it. I plan to land a set of hardening fixes early next week.

gaby:
@jayminwest Makes sense to add those too. The more secure the platform the better, specially with all the news recently of agents escaping sandboxes.

PR #820, opened 2026-08-09
fix(reap): path-limit the seeds commit to carriers git knows (warren-890a)

0xdefiant:
## The bug

`stageSeedsForCommit` path-limited its `chore(warren): seeds state` commit to a **fixed** pair of pathspecs, but `git commit --only -- <path>` is a hard error when a pathspec matches nothing in the index:

```
error: pathspec '.seeds/plans.jsonl' did not match any file(s) known to git
```

A project that ran `sd init` but never `sd plan submit` has no `plans.jsonl` in either the clone or the workspace, so **every** reap failed the `seeds_commit` stage.

The stage failure is best-effort and doesn't fail the run, but it isn't harmless: it leaves the preceding `git add -- .seeds/` staged-but-uncommitted, so the agent's seeds state never reaches the push. Nothing flags the loss, because `.seeds/` is on `BOOKKEEPING_ARTIFACT_PREFIXES` and `classifyEmptyPush` treats a `.seeds/`-only dirty tree as a deliberate no-op rather than a dropped commit.

Reproduced against real git:

```console
$ git add -- .seeds/
$ git commit --only -m 'chore(warren): seeds state' -- .seeds/issues.jsonl .seeds/plans.jsonl
error: pathspec '.seeds/plans.jsonl' did not match any file(s) known to git
$ git status --porcelain
M  .seeds/issues.jsonl        # staged, uncommitted, silently dropped from the push
```

## The fix

Filter the candidate carriers through `git ls-files` after the add and path-limit the commit to what the index actually knows. `git ls-files` never errors on an unmatched pathspec, so it is the safe way to ask the question that `commit --only` answers with a failure.

An empty result now returns cleanly instead of erroring — which also makes the gitignored-`.seeds/` opt-out that the function's own doc comment already claimed actually work.

The `git diff --cached --quiet` guard uses the same narrowed set, preserving the warren-be12 symmetry between the delta check and the commit pathspecs.

## Observed on

Runs `run_p8sawzya93x6`, `run_6s95epc8wztc`, `run_n4b6bgm3bjqc` against a project whose `.seeds/` holds `config.yaml`, `issues.jsonl`, and `templates.jsonl` — no `plans.jsonl`.

Those runs also died `finalize_failed` for an unrelated reason (a `branch_push` rejected for lack of GitHub `workflow` scope). That is an operator credential issue, not a warren bug, and is not addressed here — but it is what made the `seeds_commit` error visible in the UI.

## Tests

Two regression tests in `stage.test.ts`:

- the no-`plans.jsonl` project — asserts the commit lands, is path-limited to `.seeds/issues.jsonl`, and emits no `reap_failed`
- a gitignored `.seeds/` — asserts a clean skip with no commit and no `reap_failed`

`fakeExec` gains an `lsFiles` option (defaults to "every requested pathspec matches", so existing tests are unchanged).

`bun run check:all` — 12/12 gates pass.


0xdefiant:
Closing — opened against the wrong remote. Moving this to 0xdefiant/warren.

PR #869, opened 2026-08-14
feat(lint): add check:client-contract, proving UI calls hit routes the server serves

### LINKED ISSUE #848: Add a check:client-contract gate — prove the UI calls routes the server actually serves
## The problem

Nothing verifies that the UI's HTTP calls match the server's route table. The type half of the client/server contract is guarded (`check:wire-types`), but the path/method half — *which route a call actually hits* — has no gate. The failure mode is not hypothetical; it has recurred:

- warren-b229: SDK/UI copies of wire types drifted from the server (documented in `AGENTS.md`, "Single source of truth").
- warren-1f12: a blank run-detail page caused by the UI calling surfaces the server had deleted (the plot retirement).
- warren-e274: a UI/server response-shape mismatch in run analytics.

Each was found by a human hitting a broken page, not by CI.

## Why this gate is unusually cheap here

Both sides of the comparison are already machine-readable, verified at HEAD:

- **Server side:** `src/server/handlers/route-table.ts` is the canonical HTTP surface. Each entry carries `method` and `pattern` (e.g. `"/projects/:id/seeds/:seedId"`), and the file already exports a ready-made projection: `API_ROUTE_PATTERNS` (`route-table.ts:399`), a `{method, pattern}[]`. The docs generators already parse this file — `scripts/generate-docs.ts` exports `extractRoutes(source)` (line 39) and a `:param` pattern converter, which `scripts/generate-openapi.ts` reuses.
- **UI side:** every API call lives in one module, `src/ui/src/api/client.ts`, funneled through a single `request<T>(path, opts)` wrapper (`client.ts:93-126`) called about 32 times, plus one NDJSON streamer for the two event-stream routes. Paths are template literals where every dynamic segment is `${encodeURIComponent(param)}` (e.g. `client.ts:194-196`), so mapping a template literal to a `:param` route pattern is mechanical. Method defaults to GET with explicit `method:` in opts. No fetch or API-path construction exists anywhere else under `src/ui/src` — the `/runs/:id`-shaped strings in `App.tsx` are react-router routes, not HTTP calls.

## The work

1. Write `scripts/check-client-contract.ts`: parse the request paths and methods out of `src/ui/src/api/client.ts` (template literal → `:param` normalization), and diff against the route table (import `extractRoutes` from `scripts/generate-docs.ts` or consume `API_ROUTE_PATTERNS`). Model the script's structure and output on `scripts/generate-openapi.ts` / the other `check:*` scripts: `file:line` per finding, non-zero exit on any mismatch.
2. Report both directions: a UI call with no matching route entry fails (that's the warren-1f12 class); a route no UI call references is informational only (the CLI and SDK also consume routes — absence from the UI is not a bug).
3. Add a `check:client-contract` script to `package.json` and ride it inside the `lint` gate, next to `check:wire-types` and `check:rls` — the canonical `check:all` gate manifest is frozen, so repo-specific guards ride inside `lint` rather than taking a manifest slot (see "Quality gates" in `AGENTS.md`).
4. Write a test suite next to the script (see `scripts/*.test.ts` for the convention) covering: a matching call, a call to a missing route, a method mismatch, and the `encodeURIComponent` normalization.

## Design constraint

Keep the extraction honest rather than clever. If a path in `client.ts` is built in a way the parser cannot statically resolve, the script should fail loudly on that call site ("unparseable request path at client.ts:NNN") instead of silently skipping it — a silent skip turns the gate into false confidence. An explicit, commented skip-list in the script is acceptable for genuinely dynamic cases.

## Out of scope

- Checking the CLI or SDK request surfaces (follow-up candidates; the UI is the surface that has actually drifted).
- Request/response *body* validation — that's the wire-types gate's territory.
- Any change to `ROUTE_TABLE` itself or to `client.ts` beyond what mismatches the gate surfaces (if the new gate finds real drift at HEAD, fixing those hits is in scope and is the proof the gate works).

## Getting started

`AGENTS.md` covers setup and conventions (Bun, strict TypeScript, Biome). Run `bun run check:all` before pushing — warnings count as failures. New files under `scripts/` must stay at or under 500 lines (`check:size`).

Tracked internally as `warren-4d2d`.

luantaraschi:
Closes #848.

## Summary

- `scripts/check-client-contract.ts` parses the request paths out of `src/ui/src/api/client.ts` and diffs them against `ROUTE_TABLE`. A UI call with no matching route fails the gate. A route no UI call reaches is counted, not failed, since the CLI and the SDK consume routes too.
- The parser refuses to guess. A path it cannot resolve statically fails at that `file:line` rather than being skipped, so the gate never reports confidence it does not have.
- It rides inside `lint` next to `check-wire-types.ts` and `check-rls.ts`, and runs standalone as `bun run check:client-contract`.

## What the parser accepts

Two shapes resolve, and everything else fails loudly:

- `${encodeURIComponent(x)}` becomes `:param`.
- A trailing query-string suffix drops.

The suffix rule is derived from the source instead of allowlisted. A trailing interpolation drops only when every string it can produce is empty or starts with `?`, either inline (`${qs.length > 0 ? \`?${qs}\` : ""}`) or through a local helper (`${agentsQuery(filter)}`, whose two returns are `""` and `` `?${params.toString()}` ``). If that helper ever starts returning a path segment, it leaves the set and the call site fails. There is no skip-list to go stale.

Matching is by segment, with a `:param` on either side matching anything, which is what the router does. Route extraction reuses `extractRoutes` from `scripts/generate-docs.ts`, so the gate and the docs read the table the same way.

## The falsification test

I wanted the gate to fail on drift that actually happened rather than on a fixture I wrote. `cd1f5968` is "Fix blank run-detail page; delete plot UI surfaces (warren-1f12)". Running the scan over `client.ts` at `cd1f5968^` against today's route table reports **17 call sites**, with zero unresolvable paths in a 930-line historical file:

```
GET    /plots                          POST /plots/:param/intent
GET    /plots/needs-attention/count    POST /plots/:param/rename
POST   /plots                          POST /plots/:param/status
GET    /plots/:param                   POST /plots/:param/attachments
GET    /plots/:param/summary           DELETE /plots/:param/attachments/:param
POST   /plots/:param/sync              POST /plots/:param/attachments/:param/merge
POST   /plot-plan-runs                 POST /plots/:param/questions/:param/answer
POST   /agents/refresh                 POST /projects/:param/agents/refresh
POST   /runs/:param/messages
```

Fourteen of them are the plot surfaces that commit deleted. The other three are calls today's table does not serve either, and I did not trace when each one went away. That is the warren-1f12 shape, caught before a human opens a broken page.

At HEAD the gate passes: `33 UI calls against 43 routes, 10 routes the UI never calls`. It found no drift to fix, so this PR touches no call site.

## Changes

- `scripts/check-client-contract.ts`, 403 lines, under the 500-line budget.
- `scripts/check-client-contract.test.ts`, 19 tests: the resolver shapes, the helper rule, segment matching, method reading, the fixture-repo end to end, and the real checkout.
- `package.json`: the `check:client-contract` script, chained into `lint`.
- `AGENTS.md`: the new guard, plus a count fix. That paragraph said six guards ride inside `lint` while seven already did, since `check-merge-strategy.ts` was never listed. It now says eight and lists it.

## Test plan

- [x] `bun run lint` passes, which now includes the new gate
- [x] `tsc --noEmit` passes
- [x] `bun test scripts/check-client-contract.test.ts`: 19 pass, 0 fail
- [x] `check:size`, `check:debt`, `check:dups`, `check:deps`, `check:ci-parity`, `check:agents`, `check:prose`, `gen:docs:check`, `gen:openapi:check` all pass
- [x] Falsification: the historical `client.ts` above fails with 17 findings, HEAD passes

One thing I could not run, and I would rather say so than imply otherwise. `bun run check:all` does not start on Windows: `scripts/check-all.ts` spawns `bun` through `Bun.spawnSync`, which cannot see the `.cmd` shim, so it exits with `ENOENT: uv_spawn 'bun'` before any gate runs. I ran the manifest gate by gate instead. On `bun test scripts/` this machine reports **27 failures on a clean `main`** as well, all of them Windows-specific (`Executable not found in $PATH: "mkdir"`, the `deploy/gcp/*.sh` suites, an `EBUSY` on a temp directory). With this branch the same 27 fail and 19 more pass, which is exactly the suite this PR adds:

| tree | files | pass | fail |
| --- | ---: | ---: | ---: |
| `main` at `beca0470` | 30 | 353 | 27 |
| this branch | 31 | 372 | 27 |

CI on Linux is the real check on those.


luantaraschi:
The red `ci-postgres` here is not from this branch, and it looks like it is hitting the whole repo right now.

Every failure in the job is a query against a table that was never created:

```
error: relation "public.runs" does not exist
```

So the migrations do not reach the Postgres service, and the repo, trigger, and schema suites all fall over behind that one cause.

Same signature on your own merged PRs. #867 failed `ci-postgres` at 20:26 UTC today with that exact line, and #864 was still green at 18:52, so something in that window took it out. The diff here is one new script, its test, a `package.json` entry, and an `AGENTS.md` paragraph, and it has no Postgres path in it.

The other four checks on this branch are green: `ci`, `docker-build`, `acceptance-public`, and `enable-auto-merge` skipped because the head is a fork. I have rebased onto current `main` as well. Happy to look at the Postgres setup separately if it is still red tomorrow, but I did not want to touch it from inside this PR.

luantaraschi:
The red `ci-postgres` here comes from `main`, not from this branch.

`src/db/migrations/postgres/0039_dear_calypso.sql`, added by #867, writes the foreign key as `REFERENCES "public"."runs"("id")`. `withDb` creates a per-test schema and pins `search_path` to it, so `runs` lives in `warren_test_<hex>` and `public.runs` is not there. The constraint fails, the migration never finishes, and 149 tests go down with it, all of them postgres suites.

It is the only schema-qualified reference in that folder. The other fifteen foreign keys name the table on its own. The workflow was green at 18:50 UTC and first went red at 20:25 on #867's own branch, a minute before it merged.

This PR only touches `AGENTS.md`, `package.json` and the two `scripts/check-client-contract` files. Dropping the `"public".` qualifier looks like the whole fix, and I can send that as its own PR.

PR #924, opened 2026-08-16
fix(runs): reject known provider/model mismatches

### LINKED ISSUE #897: Dispatch accepts a modelOverride incompatible with the resolved provider — validate at dispatch time instead of dying mid-run
## The defect

`POST /runs` accepts a `modelOverride` that cannot work with the run's resolved provider, queues the run, starts it, and lets it die minutes later with `failureReason=provider_error` (or `no_model_response`) and little to go on. The mismatch is knowable at dispatch time.

Real incident: after the warren project's default agent moved to `openrouter/moonshotai/kimi-k3`, a dispatch with `modelOverride=claude-opus-4-8` (an Anthropic id, valid before the switch) was accepted, started, and died as a silent dead run.

The gap at HEAD (verified 2026-08-16): `modelOverride` is a free-text string from body to frozen frontmatter, never checked against the provider.

- `src/server/handlers/runs/dispatch.ts:117` — read with a type check only.
- `src/runs/spawn/dispatch.ts:136-141` — `effectiveProvider` and `effectiveModel` are resolved side by side (override > project defaults > agent frontmatter) with no cross-check.
- `src/registry/schema.ts:366-379` (`withProviderOverrides`) and `:115-126` — accept any non-empty string.
- The UI's model field is free text too (`src/ui/src/pages/NewRun.tsx:345-354`).

Contrast with the validation that *does* exist on the same path: the runtime id goes through `assertKnownRuntimeId` (`src/registry/schema.ts:343-359`) and rejects with a 422 naming the known ids. That is the shape this issue wants for the provider/model pair.

## Why this is a challenge, not a mechanical fix

There is no provider→known-models registry to validate against. `MODEL_TIERS` (`src/registry/builtins/model-tiers.ts:38-51`) is env-overridable defaults, not a closed set, and `provider` itself is not validated even for existence. A systemic registry is tracked separately as `warren-fb8d`; this issue should not build it.

The first cut is a **conservative shape heuristic that rejects only unambiguous mismatches**: OpenRouter ids contain a `/` (`vendor/model`); Anthropic ids are `claude-*` and never contain `/`. So `provider=openrouter` + a slashless `claude-*` id, or `provider=anthropic` + a slashed id, are certain mismatches worth a 4xx with a message naming both values. Anything ambiguous must pass — a false rejection is worse than the status quo. Deciding where that line sits, and writing the error message an operator can act on, is the design work.

## Where to implement

`src/runs/spawn/dispatch.ts:136-141` is the natural throw site: both effective values are in hand, after the full resolution chain, and **before** the run row is created at line 181. `ValidationError` from `src/core/errors.ts` is already imported there. Validating in the HTTP handler instead would miss the real incident shape, where the provider came from project defaults, not the request. Note the same check should fire however the values arrive — `providerOverride` with a frontmatter model has the same failure mode as the reverse.

Tests: the dispatch path has suites under `src/runs/spawn/` to extend; add cases for both certain-mismatch directions and for the ambiguous-passes rule.

## Scope

Out of scope: a provider registry (warren-fb8d), validating provider existence, network calls to providers at dispatch time, and enriching the late failure diagnostics (that is #896 / `warren-4001`).

## Getting started

`AGENTS.md` covers setup and conventions. Run `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-bad5`.

**Revised 2026-08-16.** The original internal report also asked, as a fallback, to "at minimum surface the provider error message on the run" — that half was already shipped before the report was filed (PR #555, commit `cefbba52`, 2026-06-30: the reap path extracts the harness's `errorMessage` into a `reap.provider_error` event). Only the dispatch-time validation remains, and it is fully live.

JulianDuerr-THM:
## Summary

- reject provider/model pairs that are unambiguously incompatible after the full override resolution chain
- fail before seed construction, run-row creation, or runtime-provider calls
- preserve open-ended provider/model support by allowing unknown, casing-variant, and ambiguous pairs

Fixes #897

## Changes

- add a focused validator for the two known mismatch shapes
- invoke it after operator override > project default > agent frontmatter resolution
- cover both rejection directions, absence of side effects, compatible pairs, and ambiguous pass-through cases

## Test plan

- [x] `bun run typecheck`
- [x] Biome check for all changed files
- [x] focused spawn tests (37 passed, 0 failed)
- [x] `bun run check:all` in the Linux verification environment (12/12 gates)


jayminwest:
Thank you @JulianDuerr-THM, great work.

PR #981, opened 2026-08-18
fix(cli): report which slot supplied the resolved credential

### LINKED ISSUE #898: The CLI silently absorbs stale credentials from a cwd .env — track and report the resolved credential source
## The defect

The warren CLI silently absorbs credentials from any `.env` file in the directory you happen to run it from, and nothing ever tells you where the credential it used came from.

The published CLI is raw bun-shebang TypeScript (`package.json` maps `warren`/`wr` to `./src/cli/main.ts`, whose first line is `#!/usr/bin/env bun`), and Bun auto-loads `.env` from the invoking cwd. So a stale `WARREN_API_TOKEN` in some repo's `.env` outranks both the token persisted by `warren login` in `~/.warren/client.json` (env > file) and a token piped to `warren login` on stdin (env > stdin). The observed symptom: every command fails `[unauthorized] invalid bearer token` immediately after a *successful* login, and it works again from `/tmp` — with zero hint that the losing credential came from `./.env`.

Verified at HEAD 2026-08-16:

- `src/cli/client.ts:68-76` — `resolveClientConfig` collapses flag / env / config-file candidates via `firstNonEmpty` and discards which source won. No mitigation for Bun's `.env` auto-load exists anywhere: `bunfig.toml` is empty of config, and no `--env-file` handling appears in `src/cli/` or `package.json`.
- `src/cli/commands/doctor-remote.ts:94` — the auth-failure hint is `"check WARREN_API_TOKEN / --token against the server's credential"`. It never mentions `~/.warren/client.json`, `warren login`, or the possibility that the token was auto-loaded from a cwd `.env`.
- `warren login`'s success output (`src/cli/commands/login.ts:63`) reports the config path it *wrote*, not the source a later command will *resolve* — so login "succeeding" proves nothing about what `warren show` will send.

## The fix

Track the winning source through the resolution chain and name it wherever auth is reported:

1. Extend `resolveClientConfig` (`src/cli/client.ts:68-76`) to return the source of each resolved value alongside the value — e.g. `tokenSource: "flag" | "env" | "config-file"` (and the same for the base URL). The precedence chain itself must not change: flags > env > config file > default is documented behavior.
2. Surface it in `warren doctor`: the `auth_valid` check (`src/cli/commands/doctor-remote.ts:81-97`) should name the resolved source in both its ok line and its failure hint, and the hint should mention `~/.warren/client.json` / `warren login` alongside `WARREN_API_TOKEN` / `--token`. A failure line that says the rejected token came from `env` is what turns this from a debugging detour into a one-glance diagnosis (the CLI can't distinguish a cwd-`.env`-injected env var from a real one — Bun loads it before the process starts — so "env" is honest and sufficient).
3. Document the footgun: the env-contract sections in `AGENTS.md` and `docs/cli-reference.md` should warn that Bun auto-loads the invoking directory's `.env`, so a repo-local `WARREN_API_TOKEN` overrides the login config. (`docs/cli-reference.md` is generated — see `bun run gen:cli-ref`; put the prose wherever the generator sources it, not in the emitted file.)

Existing tests to extend: `src/cli/client.test.ts` covers `resolveClientConfig`'s precedence (e.g. the `tok-file` case around line 47) — add assertions for the reported source. `src/cli/commands/login.test.ts` covers login flows.

## Scope

Out of scope: changing the precedence order, suppressing Bun's `.env` auto-load (`--env-file=off` for the entrypoint is a separate discussion with compatibility implications), and any server-side change.

## Getting started

`AGENTS.md` covers setup and conventions. Run `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-8807` (duplicate report: `warren-2ceb`; absorbs the hint-text remainder of `warren-c550`).

luantaraschi:
Closes #898.

Two thirds of the issue turned out to be in the tree already, so this is only what was left. Checked at `f7b8577`:

* the `auth_valid` hint in `doctor-remote.ts` already names `~/.warren/client.json`, `warren login` and the cwd `.env`.
* the env contract already carries the warning, in `AGENTS.md` and in the `WARREN_API_TOKEN` row of the generated `docs/cli-reference.md`.
* `commandFailure` in `src/cli/output.ts` already prints a warren-8807 hint on an auth rejection when the env carries a token.

What was missing is the piece all of that leans on: the resolution never recorded which slot won.

**`resolveClientConfigWithSources`** returns the merged config plus `baseUrlSource` and `tokenSource`, each one of `flag | env | config-file | default`. `resolveClientConfig` is now a one line wrapper over it, so every existing caller keeps the shape it had and the precedence is untouched. `tokenSource` is absent when no slot carried a token at all.

**`warren doctor` names the slot.** The ok line reads `admitted as operator (capabilities: ...), token from WARREN_API_TOKEN in the environment`. On a rejection the hint leads with `the rejected token came from ...` and then says what to do about that one slot, instead of listing all three candidates and leaving the operator to guess. `server_reachable` names the base URL slot the same way. With no source resolved, both fall back to the wording they have today, so nothing that builds deps by hand changes.

The CLI still cannot tell a cwd `.env` from a real environment variable, since Bun loads the file before the process starts. Both report `env`, which is the honest answer the issue asks for, and the one that explains an `[unauthorized]` right after a successful `warren login`.

**Why the production wiring sits in `doctor-remote.ts`.** `remoteDoctorDeps(env, flags)` is the factory `main.ts` calls. Putting it there keeps `main.ts` flat: that file is at 495 lines against the 500 line guard, and importing a fifth name from `./client.ts` alone wraps the import and spends six of the five lines left. The module that owns `RemoteDoctorDeps` owning how production builds them also matches the `Production wires ...` note already on those deps.

## One thing I did not do

`commandFailure` still decides from `hasEnvToken(context.env)`, so a rejected `--token` is blamed on the environment. Making it accurate means carrying the resolved source through `CliContext`, which reaches every command runner, and that is a wider change than the seam this issue describes. Say the word and I fold it into this PR.

## Tests

Ten new tests, each written before the code that makes it pass.

`src/cli/client.test.ts`, four cases on the resolver: the built in default reports `default` with no token source, each slot reports itself, a `--url` flag with an env token reports the two halves apart (`flag` and `env`), an empty string flag does not claim the slot it lost, and the merged config still equals what `resolveClientConfig` returns.

`src/cli/commands/doctor-remote.test.ts`, six cases on the report: the factory carries the slots through, the reachability line names the base URL slot, the ok line names the token slot, and a rejected token names `env`, `--token` or the config file. Before the change every one of them read the sourceless text, for example the config file case answered with the generic `check WARREN_API_TOKEN / --token or ...` hint.

## Gates

Run on this branch, Windows, Bun 1.3.14:

* `bun run lint` clean, all ten house checks plus `gen:cli-ref:check`
* `bun x tsc --noEmit` clean
* `bun run scripts/check-file-sizes.ts` clean, `main.ts` back at its original 495 lines
* `bun test src/cli`: 318 pass, 4 fail. The same 4 fail on `origin/main` on this machine (308 pass there), all file mode and sqlite path cases Windows cannot satisfy: `saveWarrenClientConfigToFile` 0600 twice, `runLogin` 0600, `resolveDbUrl` sqlite default.
* `bun test` whole suite: 5162 pass, 104 fail, against 5152 pass and 104 fail on `origin/main` on the same machine. I compared the failing test names one by one: the two sets are identical, so nothing on this branch fails that passes on the base. The 104 are the known Windows ones (file modes, sqlite paths, bwrap and sidecar probes).

This contribution was made with AI assistance. I reviewed the change and ran every command quoted above.


jayminwest:
Merged, thanks — the resolver seam and the doctor wiring are exactly the missing piece, and the test-first writeup made this easy to verify.

On your open question: yes, please keep the commandFailure generalization out of this PR. It is now tracked as warren-2d4c (thread the resolved sources through CliContext so every command blames the right slot). It is yours if you want it — the writeup in the tracker credits your offer.

Three smaller follow-ups from review are also tracked, in case you are interested: warren-4f1b (the failure-path hints do not yet use the source info, and the client-config path should come from clientConfigPath(env)), warren-3f76 (the remoteDoctorDeps test needs the WARREN_CLIENT_CONFIG sentinel guard your client.test.ts changes use), and warren-ff07 (a pre-existing empty-WARREN_BASE_URL throw that defeats flag precedence, carried through the rewrite).

Issue #980, opened 2026-08-18
Spawn-exec failure (missing docker CLI) is misclassified as no_model_response and still pushes a bookkeeping branch


jayminwest:
> **Good first challenge** — bigger than a starter issue: adds a value to the wire vocabulary and touches the reap pipeline, but every site is mapped below and there's a merged precedent to copy.

## The defect

When the runtime cannot exec the agent at all — e.g. the `docker` CLI is missing or unexecutable under `WARREN_RUNTIME=docker` — the run fails with `failureReason: "no_model_response"`, which tells the operator "the model said nothing" when the truth is "warren never started the agent". Worse, reap still runs its full finalize pipeline and **pushes a bookkeeping branch to origin** for a run that never produced any agent work.

The trace at HEAD:

1. The spawn throw (`Bun.spawn` on a missing binary) escapes before the run is marked running in the local store (`src/runtime/docker/spawn.ts:85-89` → `src/runtime/local/drive.ts:203-205`), and `driveLocalRun` converts it into a single error event with `stream: "system"` (`drive.ts:127-138`).
2. The bridge has already claimed the row `queued → running` on that first event (`src/runs/stream/bridge.ts:162-172`), so reap's only skip arm — `stateOnEntry === "queued"` → `reap.never_started_skip` (`src/runs/reap/run.ts:144-145`) — does not fire.
3. The failure classifier (`inferFailureReason`, `src/runs/reap/state.ts:57-74`) has exactly two discriminators: model output on stdout → `crashed`, a `^(?:bwrap|sandbox-exec): ` stderr line → `sandbox_failed` (`state.ts:45`). A `stream:"system"` spawn error matches neither, so the fallback at `state.ts:73` yields `no_model_response`.
4. The finalize intent is built with `push: true` unconditionally (`src/runs/reap/finalize-intent.ts:38`) — no outcome or failure-class condition exists — and the push executes at `src/runtime/local/finalize.ts:374`. (The seeds bookkeeping commit is conditional on the project having a `.seeds/` directory with a delta; the branch push is not.)

[...]

PR #991, opened 2026-08-19
fix(judge): sync run failure reasons in judge wire types with core vocabulary

rookepoole:
## Summary
Updates `RUN_FAILURE_REASONS` in `extensions/judge/src/warren-wire.ts` to include `spawn_failed` and `push_rejected_policy`, preventing `WireDriftError` when the judge parses runs that failed due to a spawn-exec error or policy-rejected push.

Addresses #980

---


jayminwest:
Thanks @rookepoole — this landed via #1021, which carries your commit (author preserved) plus the third drifted value (`no_changes`, from #992) and a drift-guard test so the judge mirror can't fall behind core again.

Issue #979, opened 2026-08-18
A repair run that commits nothing reaps as plain succeeded — and commits_ahead is structurally 0 on ref-dispatch


jayminwest:
> **Good first challenge** — bigger than a starter issue: two coupled defects, both mapped below with file-level anchors. Comfort reading a multi-stage pipeline helps.

## The defect

Warren can dispatch a "repair run" onto an existing branch (ref-dispatch: `targetBranch` = the branch to fix, e.g. a PR's head). Observed in practice: such a run made **no commits**, reaped as plain `succeeded`, and the pipeline shepherding it read "succeeded" as "fixed" — the failure survived to the next CI round. Nothing on the run record distinguishes "agent fixed it" from "agent did nothing".

Two coupled problems at HEAD:

**1. `noChanges` is computed but never lands on the run record.** Reap detects an empty push — `classifyEmptyPush` (`src/runs/reap/util.ts:225-252`) distinguishes a genuine no-op from a dropped commit, and `emitEmptyPushIfNeeded` (`src/runs/reap/pipeline.ts:245-263`) sets `state.noChanges` and emits a `reap.empty_push` event. But the terminal-state decision (`src/runs/reap/run.ts:183-219`) never consults `state.noChanges`: the run row gets `succeeded` / `failureReason: null`, and the flag survives only on two event payloads (`pipeline.ts:254-262`, `run.ts:345`). The design-intent comment at `src/core/wire.ts:128-132` documents this as deliberate, and `src/runs/reap/no-changes.test.ts:88-100` pins it. The consequence: every consumer that cares must scan the event stream — `src/plan-runs/merge-gate.ts:227-240` (`hasEmptyPushEvent`) does exactly that, and the seed-close gate reinvented the same check off `commitsAhead` (`src/runs/reap/seed-close-lifecycle.ts:72-81`, from PR #837).

[...]

PR #994, opened 2026-08-19
fix(reap): measure pre-push commitsAhead correctly on ref-dispatch runs

rookepoole:
Fixes an issue where ref-dispatch repair runs (`branch === baseBranch`) evaluated `commits_ahead` as structurally 0 because `git rev-list` was executed after `git push` against the local branch name (`baseBranch..HEAD`), which evaluates to 0 when `HEAD` is attached to `baseBranch`. `countCommitsAhead` now runs before `git push` and evaluates against `origin/<baseBranch>` when `branch === baseBranch`, ensuring productive repair runs reap as succeeded with their true `commitsAhead` count while zero-commit repair runs fail as `no_changes`.

Addresses #979

---

### DIFF
--- a/src/runs/reap/no-changes.test.ts
+++ b/src/runs/reap/no-changes.test.ts
@@ -161,6 +161,29 @@ describe("reapRun ref-dispatch zero-commit (warren-ba08)", () => {
 		});
 	});
 
+	test("ref-dispatch with commits made reaps as succeeded", async () => {
+		const { repos, runId } = await setupSeeded(null, {
+			ref: "fix/pr-head",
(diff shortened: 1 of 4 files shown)


jayminwest:
Thanks @rookepoole — merged via #1022, which carries your commit (author preserved) and reworks the measurement so the finalize stage trail keeps its push→count order (the five tests CI flagged here), the outcome-facts diff uses the same pre-push ref as the count, and an unresolvable tracking ref fails to `null` instead of falling back to the structurally-empty `base..HEAD`.

PR #1041, opened 2026-08-21
fix(cli): let --url override empty WARREN_BASE_URL

### LINKED ISSUE #1034: CLI: an empty WARREN_BASE_URL in the environment throws before --url is considered
## The defect

The CLI merges its client config as flags > env > config file > built-in default, and `firstNonEmpty` (`src/cli/client.ts:149-158`) deliberately treats an empty string as unset so an empty value can never beat a good one. But `resolveClientConfigWithSources` calls `loadWarrenClientConfigFromEnv(env)` unconditionally *before* the merge (`client.ts:119`), and that loader throws `ValidationError("WARREN_BASE_URL is set to an empty string")` (`src/client/config.ts:19-27`). The result of that call is only ever used for its default base URL (`client.ts:131`).

Reproduced at HEAD:

```
$ WARREN_CLIENT_CONFIG=/nonexistent/x.json WARREN_BASE_URL= bun src/cli/main.ts doctor --url http://127.0.0.1:9
warren: [validation_error] WARREN_BASE_URL is set to an empty string
  hint: unset WARREN_BASE_URL to fall back to http://localhost:8080
```

With the env var unset, the same command honours `--url` and reports `server_reachable: false … http://127.0.0.1:9`. The practical trigger is a `WARREN_BASE_URL=` line left in a repo's `.env` — Bun auto-loads `.env` from the invoking directory — which then defeats an explicit `--url` on every command, contrary to the documented precedence.

## Fix direction

In `resolveClientConfigWithSources`, stop calling `loadWarrenClientConfigFromEnv`. Take the default slot from the exported `DEFAULT_WARREN_BASE_URL` (`src/client/config.ts:3`) and let `firstNonEmpty` skip the empty env slot like any other. Preserve the existing failure for the case where the empty env var is the *only* candidate (no flag, no config-file URL) — throw the same `ValidationError` with the same recovery hint only when nothing higher resolved. Leave `loadWarrenClientConfigFromEnv` itself alone; the SDK's `src/client/config.test.ts` pins its contract.

Tests: add to the `resolveClientConfigWithSources` cases in `src/cli/client.test.ts` — "empty WARREN_BASE_URL loses to --url" and "empty WARREN_BASE_URL with no flag still fails with the existing error".

## Scope

Out of scope: the precedence order, the token slot, and `loadWarrenClientConfigFromEnv`'s behaviour.

## Getting started

`AGENTS.md` covers setup and conventions. `bun test src/cli/client.test.ts` runs the relevant suite; run `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-ff07`.

HostX0:
Closes #1034.

## Summary

The CLI's documented precedence is `flags > env > config file > default`, and `firstNonEmpty` already treats an empty string as unset. The merge never got a chance to apply that rule, though, because `resolveClientConfigWithSources` eagerly called `loadWarrenClientConfigFromEnv`, which throws as soon as `WARREN_BASE_URL=""` is present.

This change keeps the SDK env-loader contract untouched and resolves the CLI URL slot directly:

- an explicit `--url` wins over an empty `WARREN_BASE_URL`;
- an empty env URL can fall through to a valid config-file URL;
- if no flag, env, or config-file URL resolves, the existing `ValidationError` and recovery hint are preserved exactly;
- token precedence is unchanged and remains independent from the URL slot.

## Regression coverage

Added focused cases in `src/cli/client.test.ts` for the flag winner, the config-file fallback, and the preserved no-candidate validation error (including its recovery hint).

I was not able to run Bun locally from this execution environment, so I am not claiming local gate results; the diff is intentionally limited to the CLI resolver and its existing test suite.


jayminwest (approved):
Reviewed after updating onto current main: precedence behavior and regression coverage are correct; fresh CI is green.

PR #1044, opened 2026-08-22
fix(k8s): keep judge secret template out of kustomize

### LINKED ISSUE #1038: deploy/k8s judge kustomization renders the placeholder secrets.yaml, so apply -k clobbers the live judge-secrets with REPLACE_ME
## The defect

`deploy/k8s/extensions/judge/kustomization.yaml:17-21` lists `secrets.yaml` in `resources`. That file is a placeholder template — its own header (`deploy/k8s/extensions/judge/secrets.yaml:1-8`) says "do NOT `kubectl apply` this file", and its values are `judge-export-token: "REPLACE_ME"` (`:29`) and `openrouter-api-key: "REPLACE_ME"` (`:32`). The kustomization's own header (`:6-8`) says the Secret is created imperatively. So any overlay that includes `../../extensions/judge` renders — and on `kubectl apply -k` writes — a `judge-secrets` Secret with placeholder values over the real one. Rendering such an overlay at HEAD emits `judge-export-token: REPLACE_ME` / `openrouter-api-key: REPLACE_ME` under `kind: Secret, name: judge-secrets`. This took out the live judge deployment on 2026-08-19 (recovered by re-creating the Secret by hand).

## Fix direction

The main base has the identical shape (`deploy/k8s/base/kustomization.yaml:17` lists `secrets.yaml`) and the `gke` overlay neutralizes it with `$patch: delete` entries for each template Secret (`deploy/k8s/overlays/gke/kustomization.yaml:70-101`, with a comment saying exactly why). For the judge the smaller and more honest fix is the one the internal row asked for: drop `secrets.yaml` from the judge kustomization's `resources` so the template is docs-only, which is what both headers already claim. Adjust the header comment at `kustomization.yaml:6-8` to say the file is documentation and is not rendered.

Verify locally, no cluster needed: `kubectl kustomize deploy/k8s/extensions/judge | grep -c REPLACE_ME` → `0`, and the pvc / deployment / service still render. `scripts/deploy-trigger.test.ts` renders only the `gke-live` overlay today; a small companion test that renders the judge kustomization and asserts no `REPLACE_ME` is welcome if `kubectl` is available in the test environment — if not, the manual check is enough.

## Scope

Out of scope: the operator-local `gke-live-judge` overlay (gitignored), the base `secrets.yaml` handling, and `.github/workflows/deploy-gke.yml`.

## Getting started

`AGENTS.md` covers setup and conventions; `deploy/k8s/README.md` and `docs/RUNBOOK-K8S.md` describe the manifests. Run `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-0812`.

HostX0:
## Summary

The judge kustomization currently includes `secrets.yaml` in `resources`, even though that file is explicitly a placeholder template with `REPLACE_ME` values and its header says it must not be applied directly.

This change removes the template Secret from the rendered judge resources and clarifies that `secrets.yaml` is documentation-only. The real `judge-secrets` Secret remains operator-created, matching the existing deployment instructions and preventing `kubectl apply -k` from overwriting live credentials with placeholders.

Fixes #1038.

## Validation

- Audited the resulting diff: one file, `+2/-2`, with only the placeholder Secret resource removed and the deployment comment clarified.
- I did not run `kubectl kustomize` locally in this environment, so I am not claiming that command as executed validation.

### DIFF
--- a/deploy/k8s/extensions/judge/kustomization.yaml
+++ b/deploy/k8s/extensions/judge/kustomization.yaml
@@ -5,7 +5,8 @@
 #
 # Deploy with an overlay that pins the real image (see
 # ../../overlays/gke-live-judge, gitignored) after creating the judge
-# Secret imperatively (secrets.yaml documents the keys).
+# Secret imperatively. secrets.yaml documents the required keys but is
+# intentionally not rendered by this kustomization.
 #
 # The image (warren-ext-judge) is built + pushed by CI beside the other
 # warren images, and the release deploy rolls an EXISTING judge Deployment
@@ -15,7 +16,6 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-  - secrets.yaml
   - pvc.yaml
   - deployment.yaml
   - service.yaml


jayminwest (approved):
Reviewed: the placeholder Secret is no longer rendered, local `kubectl kustomize` validation passes, and fresh CI is green.

PR #1045, opened 2026-08-22
feat(runs): bound the provider retry and say where it stopped

### LINKED ISSUE #1031: Provider-error auto-retry: bounded multi-attempt retry, an explicit exhausted marker, and a design record for the policy
> **Good first challenge** — the retry path is one self-contained module with a thorough test suite. This extends its policy, adds two event kinds, and writes the design record it never had.

## The defect

When a run dies with `failureReason: provider_error` (the model provider failed, not the agent), warren may auto-dispatch a replacement run. That logic is `src/runs/retry/provider-retry.ts`, an observe-only `post_reap` lifecycle extension wired in `src/server/main/provider-retry-wiring.ts`. Three things about it are invisible to an operator:

1. **It retries exactly once.** `maybeRetryProviderError` returns before classification if the failed run's event stream carries a `spawn.provider_retry` marker (`provider-retry.ts:255-258`); the module header documents "the retry of a retry is never dispatched" (`:25-31`). Two consecutive transient failures on one seed therefore need a human to re-dispatch by hand. Observed 2026-08-19: a run died with `Upstream idle timeout exceeded` (classified transient), the auto-retry died the same way, then nothing.
2. **Exhaustion and skips are silent.** The bound at `:258` is a bare `return` — no log line, no run event. The non-transient skip at `:262-268` logs `provider-retry.skipped` at info level and emits no event. The only event kinds the module owns are `spawn.provider_retry`, `reap.provider_retry_dispatched`, and `reap.provider_retry_failed` (`PROVIDER_RETRY_EVENTS`, `:60-67`). Nothing on the final run says "warren considered retrying this and stopped, because X."
3. **The policy is undocumented.** `rg provider-retry docs/` finds only the plan-run *child* retry section (`docs/design/plan-run-coordinator.md:110-118`) and CHANGELOG lines. The three-tier classifier (structured `httpStatus` → upstream response body → regex over the message; `:158-176`) and both pattern lists (`:84-146`) exist only as comments in the module.

## Fix direction

Smallest coherent scope, three parts:

1. **Bounded N attempts.** Replace the boolean marker check at `:255-258` with a count against a module constant (e.g. `MAX_PROVIDER_RETRIES = 2`). Two ways to count: walk the `retryOf` chain backwards — `src/runs/dispatch-context.ts:340-365` already walks retry lineage and is the precedent — or count forwards with `repos.runs.findByRetryOf` (`src/db/repos/runs-queries.ts:241`). Keep emitting the `spawn.provider_retry` marker (it is still the lineage stamp); just stop treating its mere presence as the bound. A short fixed backoff before redispatch is optional — say in the PR whether you added one.
2. **Make the stop visible.** Add `reap.provider_retry_exhausted` (on the final run, carrying the attempt count) and `reap.provider_retry_skipped` (carrying the classifier `verdict` and message) to `PROVIDER_RETRY_EVENTS` and emit them through the existing `makeRunEventEmitter` (`:394-411`). The sibling infra-lost retry has typed skip reasons worth mirroring (`src/runs/retry/infra-lost-retry.ts:105-141`).
3. **Write it down.** Add `docs/design/provider-retry.md` covering: the trigger (`failureReason === "provider_error"`, plan-run children excluded at `:252`), the three classifier tiers and both pattern lists, the bound, the events, and how this relates to infra-lost retry and plan-run child retry. `docs/design/tier1-observation-bus.md` shows how an extension's wiring is described; `plan-run-coordinator.md`'s retry section is the register to match. Link it from `AGENTS.md` (the "Lifecycle bus" section is the natural home) — `bun run check:agents` verifies the path resolves.

Tests: `src/runs/retry/provider-retry.test.ts:412` (the single-retry bound) becomes an N-bound test — a retry of a retry *is* dispatched, retry N+1 is not and emits `exhausted`; add a `skipped`-event assertion next to the unrecognized-message case at `:406`; the fixture at `:309-341` is the base. Rendering the new event kinds in `src/ui/` is optional.

## Scope

Out of scope: *which* messages classify as transient (#1036), the retry dropping model overrides (#1039), infra-lost retry's own bound, and plan-run child retry (`src/plan-runs/retry.ts`).

## Getting started

`AGENTS.md` covers setup and conventions — especially "Lifecycle bus" and "Docs are a public surface". Run `bun test src/runs/retry` while iterating and `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-ac61`.

luantaraschi:
Closes #1031.

## Summary

- `MAX_PROVIDER_RETRIES` (2) replaces the boolean marker check. `countProviderRetries` walks the lineage backwards from the run that just failed and counts the runs carrying `spawn.provider_retry`.
- Two new event kinds make the stop visible: `reap.provider_retry_exhausted` carries `attempts` and `maxAttempts`, `reap.provider_retry_skipped` carries the classifier `verdict` and the message.
- `docs/design/provider-retry.md` is the record the policy never had, linked from the AGENTS.md lifecycle-bus section and catalogued in `docs/design/README.md`.

## Counting the stamps, not the `retryOf` hops

The issue offered two ways to count. I walk backwards, but over the event stamps rather than over the column, and two details of that walk are load bearing.

`retryOf` hops are not the count. `infra-lost-retry.ts:172` writes that column too, so an infra-lost retry that later dies on the provider would arrive already holding an attempt it never spent.

A run without the stamp does not end the walk either. An infra-lost retry sitting in the middle of a provider lineage is not a fresh start, so the walk passes through it and keeps counting. There is a test for exactly that shape: a lineage of `MAX` hops where the failed run is an infra-lost retry still holds one spent provider attempt, not two, and still has room for one more.

`parentRunId` is followed only from a run that carries the stamp, which is the pre-warren-eaa6 provider-retry row shape, so a plain `continue` clone is never walked. The walk stops at the cap, so it reads at most `MAX_PROVIDER_RETRIES` rows of the chain whatever the depth is.

## No backoff

The issue asked me to say either way. I did not add one.

The bus does not await its subscribers (`lifecycle-bus.ts:287-301` dispatches and attaches a `.catch`), so a sleep would not stall the run loop. It would live only in process memory: a control-plane restart during the wait drops the retry with nothing on the stream to say so, which is the same invisibility this issue is about. The classes that would want spacing, rate limits above all, already classify `durable`.

## One gate is still silent, on purpose

A run marked `provider_error` whose stream holds no `reap.provider_error` message returns without an event. That is missing data rather than a policy decision, and inventing a `verdict` for it would report a classifier run that never happened. The design record names it as a known silent path.

## Changes

| file | what |
| --- | --- |
| `src/runs/retry/provider-retry.ts` | `MAX_PROVIDER_RETRIES`, `countProviderRetries`, the two event kinds, the emits, module header |
| `src/runs/retry/provider-retry.test.ts` | the N-bound tests, the skipped-event assertion, lineage fixture |
| `src/runs/retry/provider-retry.classify.test.ts` | new: the classifier half, moved out |
| `docs/design/provider-retry.md` | new: the design record |
| `docs/design/README.md`, `AGENTS.md` | catalog row and the link |

The test file crossed the 500-line guard once the lineage fixture landed, so the classifier tests moved to their own file rather than the budget going up. The two halves share nothing but the module they come from.

## Test plan

- [x] `biome check .` passes
- [x] `tsc --noEmit` passes
- [x] `bun test src/runs/retry` passes: 33 tests across 3 files, up from 30
- [x] `bun run check:all`: 11 of 12 gates green, see below for the twelfth

Each new assertion was mutation-checked rather than assumed:

```
$ MAX_PROVIDER_RETRIES = 1
  (fail) dispatches the retry of a retry, which the old bound refused

$ both new emits removed
  (fail) records reap.provider_retry_skipped with the verdict that declined it
  (fail) stops at the bound and records reap.provider_retry_exhausted
```

`check:coverage` is the twelfth gate and it fails on this machine, which is Windows. It fails the same way on unmodified `main`. I compared the two failure sets rather than assuming:

```
$ bun test                              # this branch
  108 fail, 5610 tests across 561 files

$ bun test                              # main
  107 fail, 5607 tests across 561 files

  (empty)
```

The 104 distinct failing tests are identical in both, and the count differs only by the three tests this PR adds. They are all POSIX assumptions: `/data/projects` fallbacks, `chmod 0600` round-trips, sqlite file paths, and the macOS seatbelt profile. `check:size`, `check:dups`, `check:deps`, `check:debt`, `check:agents`, `check:design-docs`, `check:prose`, `check:ci-parity`, `gen:docs:check`, `gen:openapi:check`, `lint` and `typecheck` all pass locally.

I did not exercise a real provider failure against a live cluster.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -612,6 +612,9 @@ observe-only process singleton that run-lifecycle call sites emit into.
 Proof consumers are the healer (`src/healer/lifecycle.ts`) and the
 seed-close reap hook (`src/runs/reap/seed-close-lifecycle.ts`). See
 [docs/design/tier1-observation-bus.md](docs/design/tier1-observation-bus.md).
+The run-level provider retry (`src/runs/retry/provider-retry.ts`) is a
+third consumer. Its policy is in
+[docs/design/provider-retry.md](docs/design/provider-retry.md).
 
 ## Extensions
 
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -39,6 +39,7 @@ This table is the complete inventory. `Approved` means a design is coherent enou
 | [IssueTracker contract](./issue-tracker.md) | `contract` | `approved` | `shipped` | 2026-08-20 |
 | [PlanRun coordinator](./plan-run-coordinator.md) | `contract` | `approved` | `shipped` | 2026-08-01 |
 | [Preview environments](./preview-environments.md) | `contract` | `approved` | `shipped` | 2026-08-01 |
+| [Provider-error retry](./provider-retry.md) | `contract` | `approved` | `shipped` | 2026-08-22 |
 | [Runtime and supervisor](./runtime-and-supervisor.md) | `contract` | `approved` | `shipped` | 2026-08-01 |
 | [Docker runtime provider](./runtime-docker-provider.md) | `contract` | `approved` | `shipped` | 2026-08-17 |
 | [RuntimeProvider contract](./runtime-provider-contract.md) | `contract` | `approved` | `shipped` | 2026-07-09 |
--- a/docs/design/provider-retry.md
+++ b/docs/design/provider-retry.md
@@ -1,0 +1,159 @@
+# Provider-error retry
+
+**Kind:** contract
+**Design state:** approved
+**Delivery:** shipped
+**Arrived:** 2026-08-22
+**Shipped:** v0.17.0
+**Current truth:** `src/runs/retry/provider-retry.ts` and
+`src/server/main/provider-retry-wiring.ts`
+
+The run-level fallback for a model provider that fails transiently
+(warren-339d). The pi harness owns the in-stream reconnect, which warren
+cannot reach from the control plane, so when a run reaps
+`failed`/`provider_error` on a network-class message, warren dispatches a
+replacement run and records the lineage on both streams. The attempt
+bound, the exhausted and skipped events, and this record arrived with
+warren-ac61.
+
+**Grounds:** [`PHILOSOPHY.md`](../PHILOSOPHY.md) operating rule 6
+("read-only as long as possible"): the module is an observe-only Tier-1
+consumer of the [observation bus](./tier1-observation-bus.md), not a
+branch inside the reap path.
+
+---
+
+## 0. Trigger
+
+The module registers one `post_reap` hook. Every gate below fails closed,
+so any uncertainty means no retry.
+
+| Gate | Source |
+|---|---|
+| The run reaped `failed` | `envelope.payload.outcome` |
+| `failureReason === "provider_error"` | `runs.failure_reason` |
+| The run still belongs to a project | `runs.project_id` |
+| The run is not a plan-run child | `trigger` is neither `plan-run` nor `auto_plan_run` |
+| The lineage has attempts left | see [the bound](#2-the-bound) |
+| The provider's terminal error reads transient | see [classification](#1-classification) |
+
+A plan-run child is excluded because the coordinator owns child-level
+retry (warren-6de9). A run-level retry underneath it would double-dispatch
+the child.
+
+## 1. Classification
+
+`classifyProviderError` is the single discriminator, and it is pure. It
+returns `transient` (retry), `durable` (the provider rejected the request
+itself, so a retry fails the same way), or `unknown` (unrecognized, which
+fails closed). Three tiers, in order:
+
+1. **Structured `httpStatus`**, captured on the `reap.provider_error`
+   payload by warren-4001's enrichment. `>= 500` is transient, `400` to
+   `499` is durable. This tier exists because a 5xx whose prose never
+   spells out the code used to fail closed as `unknown` (warren-f8b2).
+2. **The structured `upstreamBody`**, which carries the provider's real
+   rejection text even when the harness surface message is opaque, such as
+   pi's `Provider returned error` (warren-eaa6). Classified by the prose
+   rules below; `unknown` falls through to tier 3.
+3. **The free-prose message**, by the same rules.
+
+Inside a tier, durable wins over transient, so a message that names both a
+symptom and a cause ("request failed: 401 ...") does not retry.
+
+**Durable patterns** cover auth and permission (401, 403, api key,
+forbidden), model rejection (404, model not found, `not_found_error`),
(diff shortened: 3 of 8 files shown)


jayminwest (changes requested):
Please rebase after #1052 before merge: this head still drops the original provider/model/cost-cap on retries, and raising the retry count amplifies that behavior. Also reconcile the documented “missing provider signal remains silent” path with the current exhaustion-before-signal ordering.

luantaraschi:
Rebased onto main, so #1052 is underneath now and `inheritedDispatchOverrides` carries the provider, model and cap the failed run resolved. One conflict, in the fixture's option bag, where #1052 added `frontmatter` and this branch added `priorRetries`. Both are wanted, so both stayed.

On the amplification point: the inheritance test that came with #1052 fires on the first retry, and the bound is what makes a lineage go deeper than that. So there is now a test that the overrides still ride the last retry the bound allows, at `priorRetries: MAX_PROVIDER_RETRIES - 1`.

On the ordering. You are right that the two disagreed, and the code was the side that had drifted. The bound was checked before the signal, so a run whose stream holds no `reap.provider_error` message emitted `reap.provider_retry_exhausted` once its lineage was spent, while section 3 of the design doc says that case returns without an event. The doc's reasoning also settles which side should move: a verdict there would misreport a classifier run that never happened, and an exhaustion verdict does exactly that. The signal gate runs first now. It is also the cheaper order, since the signal is already in the events the handler read and the bound reads chain rows. The doc says so in as many words, and there is a test that an exhausted lineage with no signal writes no verdict at all.

One thing I did not expect. `provider-retry.test.ts` was at 499 lines on main against the 500 ceiling, so the two new tests put it over. Rather than raise a budget that only ratchets down, the bound cases moved to `provider-retry.bound.test.ts` and the harness the two suites share moved to `provider-retry.test-fixture.ts`. Nothing in the tests changed in the move.

`bun test` is 104 failures here and 104 on `origin/main`, identical sets under `comm -3`. A first run also showed `article-ix-gate`'s "protected path hidden among many ordinary files", which passes on its own (14 of 14) and did not come back on a second full run, so I am calling it a flake under parallelism rather than something this branch did. `lint`, `typecheck`, `check:size`, `check:agents`, `check:dups`, `check:debt` and `check:ci-parity` pass.

jayminwest (approved):
Re-reviewed the updated head: retry override inheritance and missing-signal ordering are fixed, focused retry tests pass, and required CI is green.

PR #1046, opened 2026-08-22
chore(ci): report what overshot check:size on the PR

### LINKED ISSUE #1032: check:size has no merge-union heal: two green PRs can push a file over its frozen line budget and nothing tells the PR what overshot
> **Good first challenge** — a new GitHub Actions workflow cloned from an existing one, plus (optionally) a small flag on a 179-line script. No cluster, no credentials beyond the App token the sibling workflow already uses.

## The defect

`check:size` (`scripts/check-file-sizes.ts`) enforces a frozen per-file line ceiling — 500 by default, or the grandfathered number in `scripts/file-size-budgets.json` — and by design the ratchet only goes down (`check-file-sizes.ts:13-17`; the failure text at `:131` reads "refactor instead of raising the budget").

When a file sits at its ceiling and two independently-green PRs each add lines to it, the update-branch merge pushes the *union* to the PR branch, CI re-runs on the union, and `check:size` fails there — even though both sides passed alone. Observed three times on 2026-08-19/20 (PRs #1007, #1009, #1011, all on `src/runs/spawn/dispatch.ts`, then pinned at exactly 500 lines); each needed a human to hand-reflow comments on the PR branch. The bundle-size gate has a heal for exactly this class (`.github/workflows/bundle-size-autoheal.yml`); `check:size` has nothing — no workflow mentions it, and the script has no report or update mode.

The class is structurally present at HEAD. Twelve files have zero headroom — e.g. `src/core/wire.ts` 606/606, `src/ui/src/api/types.ts` 776/776, `src/runs/reap/pipeline.ts` 528/528, `src/db/schema/sqlite.ts` 564/564, and five unlisted files at exactly 500/500 (`src/runs/stream/bridge.ts`, `src/server/bridge-reconnect.ts`, `src/registry/schema.ts`, `src/runtime/k8s/log-stream.test.ts`, `src/runs/reap/preview.test.ts`) — and 32 files sit within 10 lines of their ceiling.

## Fix direction

An honest heal cannot raise a budget (that is the whole point of the ratchet), and an auto-reflow bot editing source is not a bounded action the way the bundle re-baseline is. What is actually missing is telling the PR *exactly what overshot*, without a local repro:

1. **New workflow**, e.g. `.github/workflows/size-overshoot-report.yml`, cloned from the skeleton of `bundle-size-autoheal.yml`: `workflow_run` on `CI` completed (`:26-29`), the same job-level `if` (failure, non-main branch, same repo; `:45-48`), the App token mint, the `pr-gate` step that proves an open same-repo PR (`:63-77`), and checkout of the head branch. Then run `bun run check:size`; if it passes, exit with "failure was not check:size"; if it fails, post — or update in place — **one** PR comment listing each `path: lines/budget (+N)` with the line "budgets only ratchet down; reflow or split the file", and exit 0 so the reporter never masks the real failure. `seeds-merge-autoheal.yml` is a second example of the App-token pattern if the bundle one is unclear. No existing workflow posts a PR comment, so `gh pr comment` (or `gh api`) with the App token is new ground — keep it to a single comment that is edited on re-run rather than appended.
2. **Optional companion flag** on `check-file-sizes.ts`: a non-failing `--headroom N` mode that prints files within N lines of their ceiling (the hot-files list above), so the comment — and `report:quality-metrics` (`scripts/report-quality-metrics.ts:174-190` already reads the budgets file) — can warn *before* a union lands. If you add it, extend `scripts/check-file-sizes.test.ts`.

Tests: the workflow itself has none (matching `bundle-size-autoheal.yml`); `scripts/check-file-sizes.test.ts` if the flag is added. `bun run check:ci-parity` must stay green — `scripts/ci-parity-config.json` holds the escape hatches if the new workflow needs one.

## Scope

Out of scope: raising any budget, auto-editing source on the PR branch, changes to `.github/workflows/auto-merge.yml` (CI on the union already *detects* the failure — detection is not the gap), and any warren-side dispatch-time warning.

## Getting started

`AGENTS.md` covers setup and the gate vocabulary (see "Quality gates", especially `check:size` and `check:ci-parity`). Run `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-8746`.

luantaraschi:
Closes #1032.

## Summary

- `.github/workflows/size-overshoot-report.yml`: on a failed CI run for a same-repo PR branch, re-runs `check:size` and, if that was the failure, posts one comment naming each overshoot as `path: lines/budget (+N)`. Edited in place on a re-run rather than appended. Always exits 0.
- `scripts/check-file-sizes.ts --headroom N`: a non-failing mode listing the files with `N` lines of slack or less. The report is what the comment carries.
- `AGENTS.md` documents the flag and the workflow next to the `check:size` bullet.

## Why a reporter and not a healer

The issue answers this and I agree with it, so the workflow does not touch a single source file. Two things follow from that:

It re-runs the gate instead of parsing the CI log. A failed CI run says nothing about *which* gate failed, and log scraping breaks the first time a message changes. Re-running costs one bun install and half a second, and when the failure was something else the job says so and stops without commenting.

It asks for `contents: read`, not `write`. `bundle-size-autoheal.yml` needs write because it pushes a re-baseline. This one only reads the tree and posts a comment through the App token.

## The comment

Built by running the workflow's own `run:` block locally. I lowered two budgets to make the gate fail on purpose, then executed the step verbatim out of the YAML:

````
<!-- size-overshoot-report -->

### The file-size guard failed on this branch

Budgets only ratchet down; reflow or split the file.

Two independently green PRs can each pass this gate alone and still push a shared
file past its ceiling once the update-branch merge lands the union, so the file
named here is not always one this PR edited.

```
Over budget (2):
  src/runs/reap/pipeline.ts: 528/520 (+8)
  src/core/wire.ts: 606/600 (+6)

Within 10 lines of the ceiling (30):
  src/db/schema/sqlite.ts: 564/564 (0 left)
  src/registry/schema.ts: 500/500 (0 left)
  ...
```
````

With the budgets restored, the same step prints `The file-size guard passes on this branch, so the CI failure was something else. Nothing to report.` and exits 0.

## `--headroom`

The issue called the flag optional. I added it because the overshoot alone does not tell a reviewer that the file was already at zero slack before either PR touched it, which is the whole shape of the bug.

It also confirms the issue's own numbers rather than restating them. `--headroom 10` on HEAD reports 12 files at exactly zero headroom and 32 within ten lines, which is what the issue says.

## Two things I could not verify

**The workflow has never run.** `workflow_run` jobs execute from the default branch's copy of the file, the same caveat `bundle-size-autoheal.yml` carries in its own header, so this cannot be exercised from a PR branch. What I could test, I did: the YAML parses, `check:ci-parity` stays green, and the `run:` block was extracted from the YAML and executed against both a failing and a passing tree, which is the output above.

**The App's scopes.** The comment step posts through the same installation token as `auto-merge.yml` and `bundle-size-autoheal.yml`. It needs `pull_requests: write`, which the App presumably already holds since it enables auto-merge, but I cannot read the installation's permissions from a fork. Worth a glance before merge. If it lacks the scope the step fails loudly rather than silently, and the CI failure it was reporting on is untouched either way.

## Test plan

- [x] `biome check .` passes
- [x] `tsc --noEmit` passes
- [x] `bun test scripts/check-file-sizes.test.ts`: 9 tests, up from 2
- [x] `bun run check:all`: 11 of 12 gates green, see below
- [x] the workflow YAML parses, and its `run:` block was executed locally against both branches of its own conditional

`check:coverage` is the twelfth gate. It fails on this machine, which is Windows, and it fails the same way on unmodified `main`. I compared the failure sets rather than assuming:

```
$ bun test | grep '^(fail)' | sort   # this branch
  104 distinct tests

$ bun test | grep '^(fail)' | sort   # main
  104 distinct tests

$ comm -3 <this branch> <main>
  (empty)
```

They are POSIX assumptions: `/data/projects` fallbacks, `chmod 0600` round-trips, sqlite file paths, and the macOS seatbelt profile. Nothing in this PR runs on that path.

## Deliberately not done

`report:quality-metrics` still reads only `scripts/file-size-budgets.json` (`summariseFileSizes`), so it reports the count of grandfathered entries and the largest budget without walking the tree. Wiring headroom into it means giving that summariser a filesystem walk, which is a different change from this one and belongs to whoever wants the panel to grow. The flag it would call is here and exported.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,25 @@ jobs:
         if: always()
         run: bun run report:quality-metrics
 
+      - name: Build the file-size overshoot report
+        if: failure()
+        run: |
+          set -euo pipefail
+          if bun run check:size > /dev/null 2>&1; then
+            echo "check:size passes on this branch, so the failure was something else."
+            exit 0
+          fi
+          mkdir -p size-overshoot
+          bun run scripts/check-file-sizes.ts --headroom 10 > size-overshoot/report.txt
+
+      - name: Upload the file-size overshoot report
+        if: failure()
+        uses: actions/upload-artifact@v7
+        with:
+          name: size-overshoot-report
+          path: size-overshoot/report.txt
+          if-no-files-found: ignore
+
       - name: Upload test timing artifact
         if: always()
         uses: actions/upload-artifact@v7
--- a/.github/workflows/size-overshoot-report.yml
+++ b/.github/workflows/size-overshoot-report.yml
@@ -1,0 +1,138 @@
+name: size-overshoot-report
+
+# Tells a PR exactly what tripped `check:size`, without a local repro.
+#
+# The file-size budgets only ratchet down, on purpose: the failure text says
+# "refactor instead of raising the budget". So there is no honest auto-heal for
+# this gate the way `bundle-size-autoheal.yml` re-baselines a measured bundle,
+# and editing a contributor's source on their branch is not a bounded action.
+# What is missing is the diagnosis. When a file sits at its ceiling and two
+# independently green PRs each add lines to it, the update-branch merge pushes
+# the union to the branch, CI re-runs on the union, and `check:size` fails there
+# even though both sides passed alone (PRs #1007, #1009 and #1011 on
+# src/runs/spawn/dispatch.ts, 2026-08-19/20). Each needed a human to work out
+# which file and by how much.
+#
+# This job re-runs the gate on the head branch and, if it is the one that
+# failed, posts (or edits in place) ONE comment naming each overshoot as
+# `path: lines/budget (+N)` plus the files closest to their own ceilings. It
+# always exits 0, so it never masks the real failure, and it re-runs the gate
+# rather than parsing CI logs, so a failure that was something else costs one
+# comment-free job and says so.
+#
+# Uses the same GitHub App installation token as auto-merge.yml and
+# bundle-size-autoheal.yml, minted per run from AUTO_MERGE_APP_ID +
+# AUTO_MERGE_APP_PRIVATE_KEY. The App needs `pull_requests: write` to post the
+# comment; it already needs that scope to enable auto-merge.
+# NOTE: workflow_run jobs run from the DEFAULT branch's copy of this file, so
+# this only takes effect once merged to main (warren-8746).
+
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+
+permissions:
+  contents: read
+
+# Same reasoning as bundle-size-autoheal.yml (warren-ffd2): do NOT guard on
+# `github.event.workflow_run.event == 'pull_request'`. Warren pushes agent
+# branches directly, so a PR branch's CI run often carries the `push` event and
+# a pull_request-only guard would silently drop the failure this workflow
+# exists to explain. Accept any failed same-repo CI run on a non-main branch
+# and let `pr-gate` prove there is an open same-repo PR first.
+
+jobs:
+  report:
+    runs-on: ubuntu-latest
+    if: >-
+      github.event.workflow_run.conclusion == 'failure' &&
+      github.event.workflow_run.head_branch != 'main' &&
+      github.event.workflow_run.head_repository.full_name == github.repository
+    env:
+      HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
+      MARKER: "<!-- size-overshoot-report -->"
+      HEADROOM_LINES: "10"
+    steps:
+      - name: Mint app installation token
+        id: app-token
+        uses: actions/create-github-app-token@v3.2.0
+        with:
+          app-id: ${{ vars.AUTO_MERGE_APP_ID }}
+          private-key: ${{ secrets.AUTO_MERGE_APP_PRIVATE_KEY }}
+
+      # Only report onto a branch with an open same-repo PR. A push-event CI
+      # failure on a branch with no open PR has nowhere to post, and this keeps
+      # the fork semantics intact (an open fork PR fails the head-repo
+      # job-level check before we ever get here).
+      - name: Require an open same-repo PR on the head branch
+        id: pr-gate
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          GH_REPO: ${{ github.repository }}
+          HEAD_REF: ${{ github.repository_owner }}:${{ github.event.workflow_run.head_branch }}
+        run: |
+          set -euo pipefail
+          number=$(gh api "repos/${GH_REPO}/pulls?head=${HEAD_REF}&state=open" --jq '.[0].number // empty')
+          if [ -z "$number" ]; then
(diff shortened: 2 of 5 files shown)


jayminwest (changes requested):
Blocking security issue: this workflow mints an App token, checks out branch-controlled code with credentials persisted, then executes `bun install` and `bun run check:size`. A same-repo agent branch can recover/use the installation token. Please redesign so privileged workflow_run code never executes the head branch, and use `persist-credentials: false`.

luantaraschi:
You are right, and the token was the smaller half of it. Checkout persisted the credential into `.git/config`, and then `bun install` plus `check:size` ran lifecycle scripts and a script file that the branch controls, in a job holding an App token with `pull_requests: write`. Warren pushes agent branches directly, so that is not a fork boundary away from anyone.

Redesigned rather than patched, because this job never needed the branch in the first place. It only reports.

CI builds the report on the branch now, where no credential is in reach, and uploads it as the `size-overshoot-report` artifact. The `workflow_run` job downloads that artifact and posts the comment. It has no checkout, no `setup-bun`, no `bun install` and no step that runs anything the branch wrote, so `persist-credentials` has nothing left to persist. Two things I kept on the trusted side: the PR number still comes from the API rather than from the artifact, and the report text goes into the comment fenced, with backticks neutralised and a byte cap, since the branch is what wrote it.

No artifact means `check:size` was not the failure. The job says so and exits 0, same as before.

One thing worth knowing, and I am not proposing it here: `bundle-size-autoheal.yml` has the same checkout-plus-`bun install` shape on the same token. The difference is that it has to write to the branch to re-baseline, so it cannot use this artifact split and would need a different answer. Happy to look at it separately if you want.

`check:agents`, `check:size`, `lint`, `typecheck`, `check:dups`, `gen:docs:check` and `check:ci-parity` all pass locally. Nothing in the tree references the workflow by name, so no test moved.

jayminwest (changes requested):
One remaining security issue in the updated workflow: `head_branch` is branch-controlled, and Git branch names may contain `&`/`%`. Concatenating `HEAD_REF` into `pulls?head=...&state=open` allows query-parameter injection and could target another PR comment. Pass the values as encoded GET fields instead:

```sh
number=$(gh api --method GET "repos/${GH_REPO}/pulls" \
  -f "head=${HEAD_REF}" -f "base=main" -f "state=open" \
  --jq '. [0].number // empty')\n```\n\n(The jq expression should be `.[0].number // empty`; spacing above is only review prose.) I prepared and validated this exact one-line-class fix locally; `check:ci-parity` passes.

jayminwest:
Exact safe replacement (the review body formatting inserted an unwanted space in the jq example):

```yaml
number=$(gh api --method GET "repos/${GH_REPO}/pulls" \
  -f "head=${HEAD_REF}" -f "base=main" -f "state=open" \
  --jq '.[0].number // empty')\n```

jayminwest (approved):
The query-parameter injection is fixed by passing encoded GET fields, including base=main. Re-reviewed the exact repair; CI parity passes.

PR #1047, opened 2026-08-22
feat(extensions): add the Jira tracker, warren-tracker/v1 over Jira Cloud

### LINKED ISSUE #1029: Jira tracker extension (warren-tracker/v1, out of process)
Successor to #662 (the Jira half). Tracker seed: `warren-27d9`.

### Status: buildable today, no core commit needed

v0.18.0 shipped the `IssueTracker` seam and the `RemoteTracker` bridge. An external tracker is a small HTTP service speaking **`warren-tracker/v1`**, pointed at by the per-project `tracker:` block in `.warren/config.yaml`. Warren never stores the tracker credential; the extension holds it. Issues and repo no longer need to live in the same system — Jira issues + GitLab/GitHub repo is exactly the shape this was built for.

- Protocol + deployment model: `docs/design/issue-tracker.md`
- Conformance suite + FakeTracker reference server: `extensions/tracker-conformance/` (`bun run check <url>`)
- Wire vocabulary: `src/tracker/remote/protocol.ts`

### Expected shape for Jira

- `supportsPlans: false` — use `POST /plan-runs` with an explicit ordered `issues: [...]` list (added in 0.18 for trackers without plans).
- `supportsMetadata: false` to start (or map warren's metadata keys to a Jira custom field later).
- `supportsScheduledIssues: false`, `isGitNative: false`.
- Statuses pass through as Jira's raw strings; warren normalizes them to open/closed/other at the bridge.
- Lives in `extensions/` or its own repo, holds its own Jira credential.

### Why this matters beyond Jira

`warren-tracker/v1` is **experimental until a foreign implementation survives the conformance suite unchanged** (PHILOSOPHY rule 4). Linear is the first-party first extension (ROADMAP Next 1); a Jira implementation by someone outside the project is the natural candidate for that foreign implementation. If you build one: run the suite and report what it made you change — that feedback is the point.

Contributions welcome.

luantaraschi:
Closes #1029.

## Summary

`extensions/tracker-jira/` serves warren-tracker/v1 against the Jira Cloud REST v3 API, out of process, holding its own credential. No core commit: the `IssueTracker` seam and the `RemoteTracker` bridge already carry it.

Capabilities are the shape the issue asked for, `supportsPlans` / `supportsMetadata` / `supportsScheduledIssues` / `isGitNative` all false, which is what Jira maps onto without inventing anything.

## The suite ran, and what it made me change

Nothing, on the protocol surface. It passed unchanged in both auth modes:

```
$ bun run dev-server --port 8792
$ cd ../tracker-conformance && bun run check http://127.0.0.1:8792
warren-tracker/v1 conformance: 7 case groups against http://127.0.0.1:8792
PASS, exit 0

$ bun run dev-server --port 8793 --bearer s3cret
$ bun run check http://127.0.0.1:8793
  FAIL capabilities/negotiation: GET /capabilities -> 401, want 2xx
$ bun run check http://127.0.0.1:8793 --bearer s3cret
warren-tracker/v1 conformance: 7 case groups against http://127.0.0.1:8793
PASS, exit 0
```

That is the real handler, the real client and the real mapping. Only the transport underneath is fake.

The feedback you asked for is not "it passed" though, it is where the protocol left a decision to me. Four, in rough order of how long each took:

**1. A tracker with no close verb.** Jira has only workflow transitions, and a workflow usually offers none out of a terminal status, so idempotent close cannot be expressed directly. What it forced is a good shape: close reads the issue first and returns there when Jira already considers it done, so closing twice is 200 both times and nothing is transitioned. Worth saying out loud that a protocol which had modeled close as "set status to X" would have been unimplementable here. The rule as written is the right rule.

**2. The retry contract is documented from warren's side only.** `issue-tracker.md` §5 says warren retries 429 and 5xx and does not retry other 4xx. A server author has to read that and infer "so a permanent refusal must be a 4xx". That inference is how `no_close_transition` became a 409 rather than a 502. Stating it as a rule *for servers* would save the next author the derivation.

**3. Nothing says what a server should do when its own upstream rejects it.** The protocol covers the bearer warren presents and documents 401 for that. It says nothing about the other credential. The naive answer, passing Jira's 401 straight through, tells warren its own bearer was refused and sends an operator to the wrong secret. I map it to `502 upstream_unauthorized`. This one deserves a line in the protocol, because the naive answer is actively misleading rather than merely underspecified.

**4. `issue_not_found` is the only reserved code.** A good minimum, but every other failure is a per-server invention, so two trackers will spell the same thing differently. Not urgent while warren maps everything else onto one `TrackerError`, and worth remembering before that stops being true.

No missing warren surface. The vocabulary plus one endpoint set was the whole contract, same finding the conformance package reported.

## The Jira mapping, and the one decision per row

| warren-tracker/v1 | Jira | note |
|---|---|---|
| `GET /capabilities` | nothing | answered locally, so a boot probe and a container healthcheck cost no Jira call |
| `GET /issues/{key}` | `GET /rest/api/3/issue/{key}` | `status` is Jira's raw name, never normalized here |
| `GET /issue-statuses` | `GET /rest/api/3/search/jql` over `JIRA_JQL` | every page, with a backstop that throws rather than truncating |
| `POST /issues/{key}/close` | read, then transition | see above |

`description` is an ADF tree on v3 and is flattened to text; a plain string passes through, which is what v2 and some proxies return. `blockedBy` reads the **inward** side of a link, which is the blocker; the outward side is the issue this one blocks and is deliberately not reported.

The pagination backstop throwing instead of truncating is deliberate. A silently short `/issue-statuses` map reads to warren as issues that vanished.

## What is not verified

**No call has ever reached a real Jira.** `src/fake-jira.ts` is transcribed from Atlassian's documented v3 response shapes, so it is a statement of what this tracker *expects* Jira to return, not evidence of what it does return. I have no instance to point it at, and I would rather mark that than let a green suite read as a live integration.

The blast radius is narrow and I named it in the README: the ADF shape of `description`, the `issuelinks` inward description on a site whose admin renamed the link type (there is a `JIRA_BLOCKED_BY_INWARD` knob for exactly that), and the `nextPageToken` pagination of `/search/jql`. A first live run either confirms those three or corrects them.

**Jira Server and Data Center are out of scope.** Their search endpoint is offset-paginated `/rest/api/2/search` and their auth is a PAT as a bearer. Shipping an untested second path would be a guess wearing a feature's clothes, so the README names the two functions that would change instead.

## Test plan

- [x] `biome check .` passes
- [x] `tsc --noEmit` passes, root and the extension's own config
- [x] `bun test extensions/tracker-jira`: 49 tests across 3 files
- [x] the conformance suite, both auth modes, output above
- [x] `bun run check:all`: 11 of 12 gates green, see below

The container was built and run, since CI's `docker-build` only builds the root image:

```
$ docker build -t warren-ext-tracker-jira .          # succeeds

$ docker run --rm warren-ext-tracker-jira
tracker-jira: JIRA_BASE_URL is required              # exit 2

$ docker run -d -p 8794:8080 -e JIRA_BASE_URL=https://unreachable.atlassian.invalid \
    -e JIRA_EMAIL=... -e JIRA_API_TOKEN=... -e "JIRA_JQL=project = WAR" warren-ext-tracker-jira
$ docker logs
tracker-jira listening on 8080 (site https://unreachable.atlassian.invalid, auth basic, jql project = WAR)

$ curl /capabilities
{"protocolVersion":"warren-tracker/v1","capabilities":{...all false}}

$ curl /issues/WAR-1
502 {"error":{"code":"upstream_unreachable","message":"jira unreachable: ..."}}
```

The boot line names the site, the auth mode and the query, which is the trio that goes wrong in practice, and never the credential.

`check:coverage` is the twelfth gate. It fails on this machine, which is Windows, and it fails the same way on unmodified `main`:

```
  (empty)          # 104 distinct tests either way
```

They are POSIX assumptions: `/data/projects` fallbacks, `chmod 0600` round-trips, sqlite file paths, and the macOS seatbelt profile. Coverage itself moved up rather than down: functions 93.42% against a 90.91% floor, lines 96.07% against 93.37%.

## One thing I did not touch

`issue-tracker.md` §8 says warren-tracker/v1 is experimental until a foreign implementation survives the suite unchanged. This is a candidate for that, but whether it counts is your call and not something a PR should assert about itself, so the doc is unchanged. The AGENTS.md extension count is the only core file this touches.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -624,13 +624,16 @@ imports it. `check:layers` enforces both directions, because an import
 either way compiles the extension into core and makes its removal
 breaking.
 
-Two extensions ship today. The flagship is `extensions/audit-log/`
+Three extensions ship today. The flagship is `extensions/audit-log/`
 (plan pl-116e), a collector that tails run events, normalizes them
 into an append-only audit log, and exports it over
 `GET /audit-log.jsonl`. Beside it sits `extensions/judge/`
 (plan pl-17ca), which reads finished runs, judges them against the
 15-class rubric v1, stores verdicts append-only, and exports them
-over `GET /verdicts.jsonl`.
+over `GET /verdicts.jsonl`. `extensions/tracker-jira/` (warren-27d9)
+is the third. It speaks the warren-tracker/v1 protocol against Jira Cloud and holds
+its own Jira credential. Its README carries the friction list for that
+build.
 
 `extensions/audit-log/FRICTION.md` logs every place the extension
 author had to work around a missing warren surface, and that list is
--- a/extensions/tracker-jira/.dockerignore
+++ b/extensions/tracker-jira/.dockerignore
@@ -1,0 +1,3 @@
+node_modules
+.git
+*.md
--- a/extensions/tracker-jira/.gitignore
+++ b/extensions/tracker-jira/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
--- a/extensions/tracker-jira/Dockerfile
+++ b/extensions/tracker-jira/Dockerfile
@@ -1,0 +1,46 @@
+# syntax=docker/dockerfile:1
+# @warren-ext/tracker-jira - warren-tracker/v1 over the Jira Cloud REST
+# API (warren-27d9).
+#
+# This image builds from THIS DIRECTORY ALONE (docs/design/extensions.md §2):
+#   docker build -t warren-ext-tracker-jira .
+#
+# It needs a Jira site, a credential and a query. Nothing is baked in:
+#   docker run --rm -p 8080:8080 \
+#     -e JIRA_BASE_URL=https://acme.atlassian.net \
+#     -e JIRA_EMAIL=bot@acme.example \
+#     -e JIRA_API_TOKEN=... \
+#     -e JIRA_JQL='project = WAR AND labels = warren' \
+#     warren-ext-tracker-jira
+#
+# The Jira credential arrives at runtime and is never written to disk or
+# logged. Pass it as a secret in any orchestrator that has one.
+
+# ---- deps stage: resolve node_modules from the package's OWN lockfile ----
+FROM oven/bun:1.2 AS deps
+WORKDIR /app
+COPY package.json bun.lock ./
+RUN bun install --frozen-lockfile --production
+
+# ---- runtime stage ----
+FROM oven/bun:1.2-slim
+WORKDIR /app
+
+COPY --from=deps /app/node_modules ./node_modules
+COPY package.json ./
+COPY src ./src
+
+USER bun
+
+ENV TRACKER_PORT=8080
+
+EXPOSE 8080
+
+# /capabilities is answered locally, so the probe reports whether this
+# container is up without spending a Jira call every 30 seconds. It
+# carries TRACKER_BEARER when one is set, or a protected container would
+# answer its own healthcheck with a 401.
+HEALTHCHECK --interval=30s --timeout=3s --start-period=10s --retries=3 \
+	CMD bun -e "const t=process.env.TRACKER_BEARER;fetch('http://127.0.0.1:'+(process.env.TRACKER_PORT??'8080')+'/capabilities',{headers:t?{authorization:'Bearer '+t}:{}}).then((r)=>process.exit(r.ok?0:1)).catch(()=>process.exit(1))"
+
+CMD ["bun", "run", "src/index.ts"]
--- a/extensions/tracker-jira/README.md
+++ b/extensions/tracker-jira/README.md
@@ -123,6 +123,7 @@ rather than a transient one, and warren does not retry a 4xx.
 | 401 / 403 | `502 upstream_unauthorized` | **this container's** Jira credential was rejected. Passing 401 through would send an operator to warren's bearer, the wrong secret |
 | 429 | `429 upstream_rate_limited`, `Retry-After` passed through | warren's bridge already backs off on 429 and honors the header |
 | 5xx or unreachable | `502 upstream_error` / `upstream_unreachable` | the tracker answered; the system behind it did not |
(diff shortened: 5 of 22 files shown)


jayminwest (changes requested):
One concrete protocol bug before merge: `decodeURIComponent` is outside the route try/catch blocks in `src/server.ts`, so malformed paths such as `/issues/%` throw instead of returning the JSON error envelope. Please handle decode failures and add coverage. A real Jira Cloud smoke test is also strongly recommended before calling this shipped.

luantaraschi:
Fixed. `/issues/%` threw `URIError: URI error` straight out of the handler, and so did `POST /issues/%/close`, both at the `decodeURIComponent` call above the `try`.

A decode that fails now returns `400 invalid_issue_id` through the same envelope as everything else. The helper returns `undefined` rather than throwing, so the route decides, and I added the code to the failure table in the README next to the other mappings.

Coverage is four malformed shapes (`%`, `%zz`, and the same two on the close path, plus a truncated multi-byte `%e0%a4%a`) asserting 400, the JSON content type and `error.code`, and one case that a well-formed escape still decodes and returns 200. On the previous head the first of those fails with the URIError at `server.ts:117`.

`bun test` in the extension is 65 passing, and `lint`, `typecheck` and `check:agents` are clean at the repo root.

On the real Jira Cloud smoke test: agreed, and I have not run one. I do not have a Jira Cloud tenant to point it at, so everything here is against `fake-jira.ts` and the conformance suite. I would rather say that plainly than let the PR read as though it had been exercised against the real API.

jayminwest (approved):
Re-reviewed the updated head: malformed encoded IDs now return the protocol error envelope on both routes, the 65-test extension suite and typecheck pass, and required CI is green. Real Jira Cloud smoke coverage remains a documented follow-up risk, not a code blocker.

Issue #1028, opened 2026-08-20
GitLab forge provider (in-core, WARREN_FORGE=gitlab)


jayminwest:
Successor to #662 (the GitLab half). Tracker seed: `warren-7ba8`, pre-work `warren-1b6f`.

### Decision (ROADMAP, 2026-08-20)

Forges stay **in-core**. The `Forge` registry (`src/forge/registry.ts`) grows one arm per forge; there is no remote-forge bridge or forge wire protocol. Trackers got a bridge (`warren-tracker/v1`) because they are a long tail of read-mostly integrations. Forges are three or four names with near-identical APIs, and the forge's load-bearing job is minting short-lived git credentials (`docs/design/forge-contract.md` §4), which should not gain a process hop. The contract was built for a second arm (`RepoRef.forge`, `parseRepoRef` chaining, `GitCredential.username`, capability flags), and acceptance scenario 40 proves the domain does not change when the forge swaps.

### Scope

- `src/forge/gitlab/` behind `WARREN_FORGE=gitlab`, implementing the `Forge` contract (`src/forge/contract.ts`): `parseRepoRef` for `https://<host>/group/sub/project(.git)` and `git@<host>:group/project` (self-hosted hosts, nested groups), `gitCredential` (PAT/project token, username `oauth2`), the PR methods over the Merge Requests API, `deleteBranch`, `botIdentity` via `/user`.
- `capabilities.checkRuns = false` is an acceptable first cut (the CI-fixer skips cleanly); pipelines/jobs mapping can follow.
- Tests: the forge contract suite, a recorded-fetch suite like `src/forge/github/`, and scenario 40 against the new arm.

### Pre-work (small, no-regret — `warren-1b6f`)

Three GitHub-isms still sit outside the seam and would bite any non-GitHub provider. Closing them first makes this issue provider-only:

1. `src/workspace/git/credential-env.ts` hardcodes the `x-access-token:<tok>@github.com` insteadOf rewrite — route username + host through `GitCredential.username` / `RepoRef`.
2. `src/projects/url.ts` checks `host !== "github.com"` — fold behind `forge.parseRepoRef`.
3. `GITHUB_TOKEN` / `githubToken` naming on the local dispatch path — the k8s path already reads `WARREN_GIT_TOKEN` with `GITHUB_TOKEN` as fallback; do the same locally.

### Out of scope

- One forge per deployment. Per-project forge selection (GitHub and GitLab side by side) is a separate feature nobody has asked for yet.
[...]

PR #1048, opened 2026-08-22
refactor(forge): stop the git credential from assuming GitHub

luantaraschi:
The pre-work of #1028 (`warren-1b6f`), items 1 and 3. It does not close the issue: the GitLab provider itself is the other half, and this is the part that has to land first for it to authenticate anything.

## Summary

- `gitCredentialGitEnv` takes a `GitSpawnCredential` (`username`, `secret`, `host`) instead of a bare token. The literals `x-access-token` and `github.com` are gone from it.
- `mintGitCredential` (was `mintGitCredentialSecret`) returns that credential. The ~20 sites that threaded a secret string thread the object, and `githubToken` / `gitToken` / `token` on those inputs become `gitCredential`.
- The plan-run child dispatch and the infra-lost retry stop holding a boot-read `GITHUB_TOKEN` and mint per dispatch off the boot forge, like every other spawn site.
- The `github` registry arm reads `WARREN_GIT_TOKEN` before `GITHUB_TOKEN`, the order `src/runtime/k8s/git-tokens.ts` already used.

## The one design decision

The issue says to route the username and host through `GitCredential.username` and `RepoRef`. The username does come from the forge. The host does not come from `RepoRef`: §0 says nothing outside a provider destructures a ref, and `RepoRef.key` is where the host would have to be read from.

So the host is read off the clone URL, in `credentials.ts`, which already holds both the URL and the forge. That is the same layout-only derivation `parseForgeOwnedUrl` already does next door, and it never cracks the ref.

Two consequences worth naming:

**FakeForge gets a credential with an empty host.** `fake://owner/name` has an authority that is the repo name, not a git host, so there is nothing an https rewrite could target. The credential still exists for the consumers that do not need a host (the in-pod credential endpoint, `authenticatedCloneUrl`), and only the rewrite renders nothing. That matches what actually happened before: the github.com-scoped rule was emitted against a `fake://` remote and never matched.

**The username is now FakeForge's own.** Three tests asserted `x-access-token:fake-credential` in the rendered env. They assert `fake:fake-credential` now, which is the point of the change.

## Two things this deliberately leaves alone

**The in-pod finalize wire.** `InPodFinalizeIntent.gitToken` still carries a bare secret, and the pod still injects it with GitHub's literals. Making it forge-neutral means bumping `IN_POD_FINALIZE_WIRE_VERSION` and deciding what a lagging agent image does with a v2 intent. That is your call, not a first-PR call, so the credential flattens to `.secret` at exactly two boundaries, both commented: `k8sForgeTokenWindows` in `runtime-wiring.ts` and the `POST /runs/:id/git-credential` handler.

**`src/projects/url.ts`.** Item 2 of the pre-work. Its `host !== "github.com"` check is a URL-parsing concern rather than a credential one, and folding it behind `forge.parseRepoRef` changes what `addProject` accepts and what its errors say. Separate PR, and I will open it against this branch if you want it.

## A near-miss worth reporting

`SECRET_FIELDS` in `log-redact.ts` listed `gitToken`. Renaming the field to `gitCredential` silently took the minted secret out of the redaction list, so a logged `FinalizeIntent` would have carried it through in the clear.

Nothing caught that: it is a string list, not a type. Both spellings are listed now, plus `WARREN_GIT_TOKEN`, and there is a test that names the hazard so the next rename fails loudly.

Related: `tsc` did not catch the four handler sites still passing `{ token: secret }` and `{ githubToken: secret }` either. Excess-property checking does not apply through a conditional spread, so `...(x !== undefined ? { token: x } : {})` type-checks against an input that has no `token`. Four sites were silently dropping the credential until the tests caught it.

## Test plan

- [x] `biome check .` passes
- [x] `tsc --noEmit` passes
- [x] `bun test`: the failing set is byte-identical to `main`

`check:coverage` is the twelfth. It fails on this machine, which is Windows, and fails the same way on unmodified `main`:

```
  (empty)          # 104 distinct tests either way
```

They are POSIX assumptions: `/data/projects` fallbacks, `chmod 0600` round-trips, sqlite file paths, and the macOS seatbelt profile.

`src/runtime/contract.ts` sat at exactly its frozen 510-line budget, so the new import cost a line. I reflowed the `gitCredential` doc block rather than raising the number.

New coverage, beyond the sites that only changed type:

- `gitCredentialGitEnv` renders a non-GitHub credential, and renders nothing when any of the three parts is empty
- `mintGitCredential` reads the host out of all three URL spellings warren accepts, and returns an empty host for `fake://`
- the `github` arm prefers `WARREN_GIT_TOKEN` and still falls back to `GITHUB_TOKEN`, so an existing deployment is untouched
- the redaction list covers both field spellings

I did not exercise a real K8s finalize or a real App-mode mint.

### DIFF
--- a/src/forge/credentials.test.ts
+++ b/src/forge/credentials.test.ts
@@ -47,6 +47,22 @@ describe("mintGitCredential", () => {
 		}
 	});
 
+	test("keeps a self-hosted forge port in the host it rewrites against", async () => {
+		// git treats host and host:port as different remotes, so an insteadOf
+		// rule that drops the port names a URL that never appears.
+		const forge = new GitHubForge({ token: "tok" });
+		forge.parseRepoRef = (url: string) =>
+			url.includes("acme.internal") ? { forge: "github", key: "x/y" } : null;
+		for (const url of [
+			"https://git.acme.internal:8443/x/y.git",
+			"ssh://git@git.acme.internal:2222/x/y.git",
+		]) {
+			expect((await mintGitCredential(forge, url))?.host, url).toBe(
+				url.startsWith("https") ? "git.acme.internal:8443" : "git.acme.internal:2222",
+			);
+		}
+	});
+
 	test("throws GitCredentialMintError on a non-no_credential mint failure", async () => {
 		const forge = new GitHubForge({ token: "tok" });
 		forge.gitCredential = () =>
--- a/src/forge/credentials.test.ts
+++ b/src/forge/credentials.test.ts
@@ -1,27 +1,50 @@
 import { describe, expect, test } from "bun:test";
-import { GitCredentialMintError, mintGitCredentialSecret } from "./credentials.ts";
+import { GitCredentialMintError, mintGitCredential } from "./credentials.ts";
 import { FakeForge } from "./fake/fake-forge.ts";
 import { GitHubForge } from "./github/provider.ts";
 
-describe("mintGitCredentialSecret", () => {
+describe("mintGitCredential", () => {
 	test("returns undefined for a URL the forge does not own", async () => {
 		const forge = new GitHubForge({ token: "tok" });
-		expect(await mintGitCredentialSecret(forge, "https://gitlab.com/x/y.git")).toBeUndefined();
+		expect(await mintGitCredential(forge, "https://gitlab.com/x/y.git")).toBeUndefined();
 	});
 
 	test("mints the static secret for an owned URL under PAT mode", async () => {
 		const forge = new GitHubForge({ token: "tok" });
-		expect(await mintGitCredentialSecret(forge, "https://github.com/x/y.git")).toBe("tok");
+		expect(await mintGitCredential(forge, "https://github.com/x/y.git")).toEqual({
+			username: "x-access-token",
+			secret: "tok",
+			host: "github.com",
+		});
 	});
 
 	test("maps no_credential (empty token) to anonymous git", async () => {
 		const forge = new GitHubForge({ token: "" });
-		expect(await mintGitCredentialSecret(forge, "https://github.com/x/y.git")).toBeUndefined();
+		expect(await mintGitCredential(forge, "https://github.com/x/y.git")).toBeUndefined();
 	});
 
-	test("mints the FakeForge credential for a fake URL", async () => {
+	test("mints the FakeForge credential, with no host to rewrite against", async () => {
+		// FakeForge OWNS fake://repo, but its authority is the repo name rather
+		// than a git host, so there is no https remote an insteadOf rule could
+		// name. The credential still exists for the consumers that do not need
+		// one; only the rewrite renders nothing (warren-1b6f).
 		const forge = new FakeForge();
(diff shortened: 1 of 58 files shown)


jayminwest (changes requested):
Credential-neutrality blockers found: (1) `outcome-facts.ts` and `pr-context.ts` pass only `.secret` to `authenticatedCloneUrl`, which still hardcodes `x-access-token`, dropping provider usernames such as GitLab oauth2; (2) `WARREN_GIT_TOKEN ?? GITHUB_TOKEN` lets an explicitly empty neutral token suppress a valid legacy token; (3) `URL.hostname` drops self-hosted forge ports from the rewrite authority. Please preserve the full credential/authority and add non-GitHub tests.

luantaraschi:
All three reproduce, and all three are fixed.

**The username.** `authenticatedCloneUrl` built `x-access-token:<secret>@host` from a literal, and the two reap call sites handed it `.secret` alone, so a GitLab credential that mints `oauth2` authenticated as the wrong user:

```
authenticatedCloneUrl("https://gitlab.com/acme/repo.git", "glpat-xyz")
  -> https://x-access-token:glpat-xyz@gitlab.com/acme/repo.git
```

Rather than add a third argument that a caller can forget again, it now takes the credential:

```ts
authenticatedCloneUrl(repoUrl: string, credential?: GitCloneCredential): string
```

Passing `.secret` alone is a type error now, which is how I found the other three call sites I had not looked at. Two of them are the init and finalize containers, where a token really does arrive with no username, so those go through `bareTokenCredential`, and that is the one place the `x-access-token` assumption is written down and named.

**The authority.** `remoteHost` used `parsed.hostname`, so `https://git.acme.internal:8443/o/r.git` produced `git.acme.internal` and the `insteadOf` rule named a remote that never appears. It reads `parsed.host` now, and the ssh spelling with a port is covered too.

**The token fallback.** `??` only falls through on undefined, so `WARREN_GIT_TOKEN=""` shadowed a valid `GITHUB_TOKEN`:

```
??  -> ""
||  -> "legacy"
```

`firstToken` now skips blanks and trims, which is what `normalizeToken` in `src/runtime/k8s/git-tokens.ts` already did. The comment beside that line claimed parity with it and did not have it, so this makes the comment true.

Non-GitHub coverage as requested: `oauth2` for GitLab and `x-token-auth` for Bitbucket in the URL helper, a self-hosted forge on 8443 and on ssh 2222 in the host reader, and both empty and whitespace neutral tokens in the registry.

`bun test` is 104 failures on this branch and 104 on `origin/main`, and the two sets are identical under `comm -3`. They are all the POSIX assumptions this Windows box trips. `lint`, `typecheck`, `check:size`, `check:agents`, `check:dups` and `check:debt` pass.

jayminwest (approved):
Re-reviewed the refreshed head after merging current main: provider usernames and authorities are preserved, blank neutral-token fallback is correct, non-GitHub coverage passes, and required CI is green. The documented in-pod v1 credential wire remains follow-up scope under #1028.

PR #1049, opened 2026-08-22
test(cli): stop the doctor-remote test reading the developer's real client.json

luantaraschi:
Closes #1033.

## Summary

`remoteDoctorDeps` runs the real resolution chain, so an env that names no `WARREN_CLIENT_CONFIG` falls through to `~/.warren/client.json`. The test at `:92` passes an env with only `WARREN_BASE_URL` and `WARREN_API_TOKEN`, so a corrupt or hand-edited file on the developer's machine fails it for a reason unrelated to the code under test.

`src/cli/client.test.ts` already solves this. The same `NO_CONFIG_FILE` constant and the same comment land here, spread into the env the test passes.

One test, nine lines. `remoteDoctorDeps` and the resolution chain are untouched, and the hint text has its own issue.

## Reproduced, then fixed

```
$ mkdir -p <scratch>/.warren && printf '{not json' > <scratch>/.warren/client.json

$ HOME=<scratch> bun test src/cli/commands/doctor-remote.test.ts     # before
  ValidationError: client config file <scratch>/.warren/client.json is not valid JSON
      at loadWarrenClientConfigFromFile (src/client/config-file.ts:48:4)
      at resolveClientConfigWithSources (src/cli/client.ts:140:19)
      at resolveWarrenClientWithSources (src/cli/client.ts:108:19)
      at <anonymous> (src/cli/commands/doctor-remote.test.ts:93:16)
  (fail) remoteDoctorDeps carries the slots the resolution picked (warren-8807)
  10 pass, 1 fail

$ HOME=<scratch> bun test src/cli/commands/doctor-remote.test.ts     # after
  11 pass, 0 fail

$ bun test src/cli/commands/doctor-remote.test.ts                    # clean HOME, after
  11 pass, 0 fail
```

That matches the reproduction in the issue exactly, including the counts. On this machine, which is Windows, the scratch HOME is set through `USERPROFILE`, which is what `os.homedir()` reads there.

## Test plan

- [x] `biome check .` passes
- [x] `tsc --noEmit` passes
- [x] `bun test src/cli/commands/doctor-remote.test.ts` passes with both a clean and a corrupt HOME
- [x] `bun run check:size`, `check:agents`, `check:debt`, `check:dups` pass

[...]

Issue #1033, opened 2026-08-20
doctor-remote.test.ts: the remoteDoctorDeps test reads the developer's real ~/.warren/client.json


jayminwest:
## The defect

The test "remoteDoctorDeps carries the slots the resolution picked" in `src/cli/commands/doctor-remote.test.ts:92-100` calls `remoteDoctorDeps(env, {})` with an env that sets only `WARREN_BASE_URL` and `WARREN_API_TOKEN`. `remoteDoctorDeps` (`src/cli/commands/doctor-remote.ts:60-62`) runs the real resolution chain — `resolveWarrenClientWithSources` → `resolveClientConfigWithSources` (`src/cli/client.ts:115-120`) → `loadWarrenClientConfigFromFile(env)` — and with no `WARREN_CLIENT_CONFIG` in that env, `clientConfigPath` (`src/client/config-file.ts:26-30`) falls through to the developer's real `~/.warren/client.json`. If that file is corrupt or hand-edited, `config-file.ts:43-49` throws `ValidationError` and the test fails for a reason that has nothing to do with the code under test.

Reproduced at HEAD: with `HOME` pointed at a scratch directory whose `.warren/client.json` contains `{not json`, `bun test src/cli/commands/doctor-remote.test.ts` → 10 pass, 1 fail (`ValidationError: client config file .../.warren/client.json is not valid JSON`). With a clean `HOME` → 11 pass.

## Fix direction

`src/cli/client.test.ts:12-15` already solves this for the resolution tests: a `NO_CONFIG_FILE` constant that points `WARREN_CLIENT_CONFIG` at an impossible path under `tmpdir()`, with a comment explaining the hazard. Copy that constant (and its comment) into `doctor-remote.test.ts` and spread it into the env passed at `:93-96`. One test, a handful of lines.

## Scope

Only that test. Do not change `remoteDoctorDeps` or the resolution chain. The doctor's *hint text* has its own separate issue.

## Getting started

`AGENTS.md` covers setup and conventions. `bun test src/cli/commands/doctor-remote.test.ts` runs just this suite; run `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-3f76`.

luantaraschi:
Taking this one.

PR #1050, opened 2026-08-22
fix(cli): let the doctor failure paths read the credential source they already have

### LINKED ISSUE #1035: warren doctor: the failure-path hints ignore the credential-source info it already has
## The defect

PR #981 taught the CLI which slot — `--url`/`--token` flag, environment variable, the `~/.warren/client.json` config file, or the built-in default — supplied the token and base URL, and `warren doctor` names that slot on its *ok* lines. The failure paths, where the information actually matters, still ignore it. At HEAD in `src/cli/commands/doctor-remote.ts`:

1. The `server_reachable` failure hint (`:90`) is the fixed string "check WARREN_BASE_URL / --url and that `warren serve` is running". `deps.baseUrlSource` is only read on the ok line (`:78-83`); `baseUrlSourceLabel` (`:114-127`) is never consulted on failure. A stale URL from the config file sends the operator to two slots that did not supply it.
2. The env-token 401 arm (`:153-154`) tells only the stale-`.env` story. Before #981 the hint also said to check the token against the server's credential (`git show 66879c4f^:src/cli/commands/doctor-remote.ts`); the flag arm (`:152`) still does, the env arm lost it.
3. `tokenSource === undefined` — no token anywhere — is a normal production state (`src/cli/client.ts:90-92` omits the field when no slot carried a token, and `src/cli/main.ts:266` wires `remoteDoctorDeps` directly), but `authFailureHint` sends it to `UNSOURCED_AUTH_HINT` (`:157-158`, text at `:110-111`), which tells the operator to check three slots that are all empty instead of "no token configured; run `warren login`". The doc comment on `RemoteDoctorDeps.tokenSource` (`:40-44`) describes the absent case as "the caller did not resolve one", i.e. a test-only state, which is wrong.
4. `~/.warren/client.json` is hard-coded four times (`:111`, `:137`, `:154`, `:156`). `clientConfigPath(env)` (`src/client/config-file.ts:26-30`) already resolves the real path and honours `WARREN_CLIENT_CONFIG`; nothing under `src/cli/` imports it.

Also stale: the doc comment on `RemoteDoctorDeps.client` (`:34`) says production wires `resolveWarrenClient(context.env, flags)`; production wires `remoteDoctorDeps` → `resolveWarrenClientWithSources` (`:60-62`).

## Fix direction

- Give the reachability failure arm a hint that switches on `baseUrlSourceLabel(deps.baseUrlSource)`, mirroring the shape of `authFailureHint` (`:149-160`).
- Add an `undefined` arm to `authFailureHint` ("no token configured; run `warren login`, or pass --token / WARREN_API_TOKEN"), fix the doc comment at `:40-44`, and restore the "check it against the server's credential" clause on the env arm.
- Thread `env` into the label/hint helpers (or pass the resolved path) and build the path with `clientConfigPath(env)` — `config-file.ts:40` and `:84` show the call.
- Fix the `:34` doc comment.

Tests, all in `src/cli/commands/doctor-remote.test.ts`: `:69-77` (reachability hint, currently asserts `toContain("WARREN_BASE_URL")`), `:79-90` (the unsourced 401 — becomes the no-token case), `:131-141` (env arm — add the credential-check assertion), `:153-162` (config-file arm — the path should follow `WARREN_CLIENT_CONFIG`), plus new cases per base-URL source.

## Scope

In scope: `doctor-remote.ts` and its test only. Out of scope: `commandFailure` in `src/cli/output.ts:221-235`, which has the same blind spot for every *other* command — that is the larger follow-up in #1030.

## Getting started

`AGENTS.md` covers setup and conventions. `bun test src/cli/commands/doctor-remote.test.ts` runs the relevant suite; run `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-4f1b`.

luantaraschi:
Closes #1035.

## Summary

All four items. #1040 already moved `authFailureHint` into `src/cli/auth-hints.ts` and fixed the hard-coded path there, so items 2, 3 and 4 land in that file plus `tokenSourceLabel`, not at the line numbers in the issue.

| # | Was | Is |
|---|---|---|
| 1 | `server_reachable` failure hint was a fixed "check WARREN_BASE_URL / --url" | names the slot that supplied the URL, and says so when it is the built-in default |
| 2 | env-token 401 arm told only the stale-`.env` story | carries "check it against the server's credential", the way the flag arm still did |
| 3 | absent `tokenSource` sent the operator to check three slots | says no token is configured and points at `warren login` |
| 4 | `tokenSourceLabel` asserted `~/.warren/client.json` | prints the path `clientConfigPath` resolves |

## Item 3, and why absent really does mean empty

The issue is right that this is a normal production state, and it is worth writing down why rather than trusting it. `resolveClientConfigWithSources` picks the token with `firstNonEmpty` over flag, env and config file, with **no default arm**, unlike the base URL. So it omits `tokenSource` exactly when all three are empty. Both callers resolve for real: `remoteDoctorDeps` wires straight through, and `resolveCommandClient` sets it on the context whenever it exists.

So the old text described a search the operator had already lost. It listed three slots and asked them to look, when the answer was that there is nothing to find.

The doc comment on `RemoteDoctorDeps.tokenSource` said the absent case meant "the caller did not resolve one", a test-only state. That is now written as what it is.

## Both surfaces at once

`commandFailure` in `output.ts` reads the same `authFailureHint`, so items 2 and 3 land on the catch-tail of every other command with no second edit. That was the point of hoisting it in #1040.

## Test plan

- [x] `biome check .` passes
- [x] `tsc --noEmit` passes
- [x] `bun test src/cli/commands/doctor-remote.test.ts src/cli/auth-hints.test.ts`: 20 pass, up from 15

Every new assertion was checked against the old code rather than assumed. With the two source files stashed and the tests kept:

```
(fail) authFailureHint > tells the env arm to check the token, the way the flag arm does
(fail) authFailureHint > reads an absent source as no token at all, not as a search to run
(fail) runRemoteDoctor > the unreachable hint names the slot the base URL came from
(fail) runRemoteDoctor > the unreachable hint says so when no slot supplied a base URL
(fail) runRemoteDoctor > the auth line names the resolved config path, not the default
15 pass, 5 fail
```

`check:coverage` is the twelfth gate. It fails on this machine, which is Windows, and fails the same way on unmodified `main`: `comm -3` between the two failure sets is empty, 104 distinct tests either way. They are POSIX assumptions (`/data/projects` fallbacks, `chmod 0600` round-trips, sqlite file paths, the macOS seatbelt profile).

## One overlap to know about

#1049 (issue #1033) also touches `doctor-remote.test.ts`, and adds the same two `node:os` / `node:path` imports at the top. Whichever merges second wants a one-line import rebase. I branched both from `main` on purpose so neither blocks the other, and I will rebase whichever loses the race.

I did not exercise a real unreachable server or a real 401 against a live warren.

### DIFF
--- a/src/cli/auth-hints.test.ts
+++ b/src/cli/auth-hints.test.ts
@@ -17,6 +17,16 @@ describe("authFailureHint", () => {
 		expect(hint).toContain(".env");
 	});
 
+	test("tells the env arm to check the token, the way the flag arm does", () => {
+		// warren-4f1b: the env arm told only the stale-`.env` story and dropped
+		// the advice the flag arm still carried.
+		for (const source of ["flag", "env"] as const) {
+			expect(authFailureHint(source, {}), source).toContain(
+				"check it against the server's credential",
+			);
+		}
+	});
+
 	test("blames the config file and points back at warren login", () => {
 		const hint = authFailureHint("config-file", CONFIG_AT);
 		expect(hint).toContain("came from the client config file");
@@ -29,10 +39,15 @@ describe("authFailureHint", () => {
 		expect(authFailureHint(undefined, CONFIG_AT)).toContain(CONFIG_AT.WARREN_CLIENT_CONFIG);
 	});
 
-	test("names every slot when no source was resolved", () => {
+	test("reads an absent source as no token at all, not as a search to run", () => {
+		// warren-4f1b: absent means every slot is empty, which is an ordinary
+		// production state. Sending the operator to check three empty slots
+		// described a search they had already lost.
 		const hint = authFailureHint(undefined, CONFIG_AT);
+		expect(hint).toContain("no token is configured");
+		expect(hint).toContain("are all empty");
+		expect(hint).toContain("warren login");
 		expect(hint).toContain("WARREN_API_TOKEN");
 		expect(hint).toContain("--token");
-		expect(hint).toContain("warren login");
 	});
 });
--- a/src/cli/auth-hints.ts
+++ b/src/cli/auth-hints.ts
@@ -20,9 +20,12 @@ import type { EnvLike } from "./output.ts";
  * cannot see: Bun loads the file before the process starts, so a stale token
  * in the repo they happen to be standing in outranks what `warren login` saved.
  *
- * An absent source still names every slot. Both callers now resolve one for
- * real, so absent means "no token anywhere"; #1035 owns rewording that arm,
- * and lands it on both surfaces at once.
+ * An absent source means NO TOKEN AT ALL, which is a normal production
+ * state rather than an unresolved caller (warren-4f1b). `resolveClientConfig`
+ * picks the token from flag, env or config file with no default arm, so it
+ * omits the source exactly when all three are empty, and both callers
+ * resolve one for real. Sending that operator to check three empty slots
+ * described a search they had already lost; the answer is `warren login`.
  */
 export function authFailureHint(source: ClientConfigSource | undefined, env: EnvLike): string {
 	const configFile = clientConfigPath(env);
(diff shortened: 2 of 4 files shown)


jayminwest (approved):
Reviewed after #1049 update: diagnostic source changes and combined test changes are sound; fresh CI is green.

PR #1051, opened 2026-08-22
fix(acceptance): drop the retired POST /agents/refresh from scenarios 19, 24 and 35

### LINKED ISSUE #1037: Acceptance scenarios 19/24/35 still call the retired POST /agents/refresh (405)
## The defect

`POST /agents/refresh` was deleted in plan pl-3a79 (2026-07-28). The route table (`src/server/handlers/route-table.ts:148-149`) serves only `GET /agents` and `GET /agents/:name`, so the router answers `405 method_not_allowed`. Three end-to-end acceptance scenarios still call it and fail on that line before testing anything:

- `scripts/acceptance/scenarios/24-preview-node-runtime.ts:174`
- `scripts/acceptance/scenarios/19-warren-on-postgres.ts:159`
- `scripts/acceptance/scenarios/35-ci-fixer-roundtrip.ts:111`

Observed on unmodified `main`: `bun run scripts/acceptance/run.ts --only 24` → `POST /agents/refresh: expected 200, got 405 method_not_allowed`. None of the three fails in CI, which is why it went unnoticed: 19 and 24 are not in the `acceptance:nightly` list (`package.json:96`), and 35 skips early unless `GITHUB_TOKEN` + `WARREN_ACCEPT_CI_FIXER_REPO` are set.

The call was how the old agent-library flow registered the `stub-shell` test agent. That agent now arrives at boot via `WARREN_SEED_AGENTS_FILE` (`scripts/acceptance/run.ts:218-221`, `scripts/acceptance/lib/fixtures.ts:175-194`, passed through to scenario-owned boots by `scripts/acceptance/lib/inproc.ts`), so the refresh is simply unnecessary.

## Fix direction

The sibling scenarios already show both shapes. `20-preview.ts:170` just drops the call behind a one-line comment; `10-reap-seeds-roundtrip.ts:102-109` replaces it with `GET /agents/<stubAgentName>` expecting 200, which proves boot-seeding landed before spawning against the agent. Use the GET probe in all three.

Verifying: the edit is covered by `bun run typecheck` / `bun run lint` on any OS. Actually running the scenarios needs Linux with bubblewrap (24 skips on macOS at its darwin guard, `24-preview-node-runtime.ts:126`); 19 additionally needs Postgres and 35 the GitHub env above. It is fine to send the PR with only the static gates green and say so.

## Scope

Only the three `expectStatus` lines. **Revised 2026-08-20:** the internal row also reported scenario 20 leaving its run stuck in `running` past the 30 s terminal timeout. That could not be reproduced off-Linux (20 skips on darwin at `20-preview.ts:117`) and nothing in the dispatch path since the report explains it, so it stays an internal follow-up and is not part of this issue.

## Getting started

`AGENTS.md` covers setup and conventions; the "Acceptance harness" section describes `scripts/acceptance/`. Run `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-deac`.

luantaraschi:
Closes #1037.

## Summary

One line per scenario, in 19, 24 and 35. `await http.expectStatus("POST", "/agents/refresh", 200)` becomes the `GET /agents/<stubAgentName>` probe, which is what scenarios 05, 07, 09 and 10 already do.

The GET rather than the bare deletion, because the probe keeps what the retired POST was standing in for: proof that boot seeding through `WARREN_SEED_AGENTS_FILE` landed before the scenario spawns against that agent. The comment is copied verbatim from those four sites, so all seven now read identically.

## The route really is gone

Checked against the served table rather than taken from the issue:

```
$ bun -e 'import { API_ROUTE_POLICIES } from "./src/server/handlers/route-table.ts"; ...'
rotas /agents servidas hoje:
  GET /agents
  GET /agents/:name
POST /agents/refresh existe? false
```

A known path with the wrong method is 405 `method_not_allowed` (`src/server/errors.ts:196`, pinned by `src/server/server.test.ts:224`), which is the failure the issue reports.

After the change, no `expectStatus("POST", "/agents/refresh"` call remains anywhere under `scripts/`. The surviving matches for that string are all comments explaining the deletion.

## What I could not run, and why

The issue says it is fine to send this with only the static gates green, so this is that PR, and here is exactly how far I got.

I tried scenario 24 on this machine and the harness stops before the scenario:

```
$ bun run scripts/acceptance/run.ts --only 24
[acceptance] acceptance: mode=in-proc tmp=...
acceptance: harness boot failed:
Executable not found in $PATH: "cn"
    at buildCanopyRepo (scripts/acceptance/lib/fixtures.ts:243:8)
```

That is the canopy CLI the fixture builder shells to. Even with it installed, this is Windows: 24 guards on darwin at `24-preview-node-runtime.ts:126` and needs Linux with bubblewrap, 19 needs Postgres, and 35 needs `GITHUB_TOKEN` plus `WARREN_ACCEPT_CI_FIXER_REPO`.

So the reproduction in the issue is the maintainer's, not mine, and I am not claiming otherwise. What I verified myself is the static half above: the route is absent from the table, the router 405s a known path on the wrong method, and the call sites are gone.

## Test plan

- [x] `biome check .` passes
- [x] `tsc --noEmit` passes
- [x] `bun run check:ci-parity` passes
- [x] `bun run check:agents`, `check:size`, `check:debt`, `check:dups` pass
- [ ] the three scenarios actually run: not possible here, see above

`check:coverage` fails on this machine, which is Windows, and fails the same way on unmodified `main`: 104 distinct tests that assume POSIX. This PR touches three files under `scripts/acceptance/scenarios/`, none of which the unit suite executes.

### DIFF
--- a/scripts/acceptance/scenarios/19-warren-on-postgres.ts
+++ b/scripts/acceptance/scenarios/19-warren-on-postgres.ts
@@ -156,7 +156,14 @@ export const scenario: Scenario = {
 			// and exercises the dialect-aware ping path on pg.
 			await assertDbReachableOnPostgres(http);
 
-			await http.expectStatus("POST", "/agents/refresh", 200);
+			// The stub agent was seeded at boot via WARREN_SEED_AGENTS_FILE
+			// (warren-e376); POST /agents/refresh is deleted (pl-3a79). GET it
+			// to prove boot seeding landed before spawning against it.
+			await http.expectStatus(
+				"GET",
+				`/agents/${encodeURIComponent(ctx.fixtures.stubAgentName)}`,
+				200,
+			);
 			const project = await ensureProject(http, ctx.fixtures.sampleProjectGitUrl);
 
 			const created = await http.expectJson<CreateRunResponse>("POST", "/runs", 201, {
--- a/scripts/acceptance/scenarios/24-preview-node-runtime.ts
+++ b/scripts/acceptance/scenarios/24-preview-node-runtime.ts
@@ -171,7 +171,14 @@ async function runNodePreview(ctx: ScenarioCtx): Promise<void> {
 		ctx.logger.info(`scenario-24: warren ready at ${handle.warrenUrl}`);
 
 		const http = new WarrenHttp({ baseUrl: handle.warrenUrl, token: handle.token });
-		await http.expectStatus("POST", "/agents/refresh", 200);
+		// The stub agent was seeded at boot via WARREN_SEED_AGENTS_FILE
+		// (warren-e376); POST /agents/refresh is deleted (pl-3a79). GET it
+		// to prove boot seeding landed before spawning against it.
+		await http.expectStatus(
+			"GET",
+			`/agents/${encodeURIComponent(ctx.fixtures.stubAgentName)}`,
+			200,
+		);
 		const project = await ensureProject(http, sample.gitUrl);
 
 		const created = await http.expectJson<CreateRunResponse>("POST", "/runs", 201, {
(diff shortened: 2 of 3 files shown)


jayminwest (approved):
Reviewed after updating onto current main: the three acceptance probes now use the supported GET route and fresh CI is green.

PR #1052, opened 2026-08-22
fix(runs): keep the provider, model and cap when a run is retried

### LINKED ISSUE #1039: Provider-error auto-retry drops providerOverride/modelOverride — the retry runs on the agent-default model
## The defect

When warren auto-retries a transient provider failure, `dispatchProviderRetry` (`src/runs/retry/provider-retry.ts:281-337`) builds the replacement `spawnRun` input from a narrowed pick of the failed run (`:284-292`: id, agentName, projectId, prompt, trigger, seedId, mode) and passes agent / project / prompt / trigger / `dispatchOrigin` / seedId / mode / `parentRunId` / `cloneKind` / `retryOf` (`:303-325`). It never passes `providerOverride`, `modelOverride`, or `maxCostUsdOverride`, all of which `SpawnRunInput` accepts (`src/runs/spawn/types.ts:100-113`).

`spawnRun` then resolves `input.providerOverride ?? projectDefaults.defaultProvider` and `input.modelOverride ?? projectDefaults.defaultModel` (`src/runs/spawn/dispatch.ts:125-129`) and folds the result onto the agent's frontmatter (`:144-151`). So a run dispatched as `openrouter / deepseek/deepseek-v4-pro-0813` retries as `openrouter / moonshotai/kimi-k3` (the `pi` built-in's default) — the provider only "survives" because the project default happens to coincide. Observed twice on 2026-08-20 (run_g0xkk7bm2c33 → run_c977hphsyxam, run_90cbqb3ryn2q → run_2gzqnxhep8dj). An operator-selected model silently swapping mid-lineage is the bug whether the swap lands on a cheaper model or a broken one.

The sibling infra-lost retry (`buildRetrySpawnInput`, `src/runs/retry/infra-lost-retry.ts:157-193`) copies `ref` / `baseCommit` / `targetBranch` / remaining cost cap but has the same provider/model omission.

## Fix direction

The HTTP clone path already does this right: `resolveCloneDefaults` in `src/server/handlers/runs/dispatch.ts:50-74` reads the parent's *effective* provider / model / cap back off its frozen `renderedAgentJson.frontmatter` with `readProviderFrontmatter` and `readMaxCostUsd`, and passes them as overrides. Do the same in `dispatchProviderRetry`: widen the `run` pick to include `renderedAgentJson`, read the frontmatter, and pass `providerOverride` / `modelOverride` / `maxCostUsdOverride` into `spawnRunFn`. Apply the same to `buildRetrySpawnInput` in `infra-lost-retry.ts`. A small shared helper ("retry-spawn fields from a failed run") used by both modules is the tidy version — `bun run check:dups` will nudge you that way if the two copies grow.

Tests: extend the fixture at `src/runs/retry/provider-retry.test.ts:309-341` so the failed run's `renderedAgentJson.frontmatter` carries a provider and model, then assert `call.providerOverride` / `call.modelOverride` on the captured spawn; mirror in `src/runs/retry/infra-lost-retry.test.ts` beside its existing spawn-input assertions.

## Scope

Out of scope: retry policy — how many attempts (#1031) and which messages count as transient (#1036) — and `retryOf` lineage, which is already stamped at `:325` since PR #985.

## Getting started

`AGENTS.md` covers setup and conventions. `bun test src/runs/retry` runs both retry suites; run `bun run check:all` before pushing — warnings count as failures.

Tracked internally as `warren-0d80`.

luantaraschi:
Closes #1039.

## Summary

- `src/runs/retry/inherited-overrides.ts`: reads `providerOverride` / `modelOverride` / `maxCostUsdOverride` off a failed run's frozen `renderedAgentJson.frontmatter`, the same three slots `resolveCloneDefaults` already reads by hand on the HTTP clone path.
- `dispatchProviderRetry` spreads all three into `spawnRun`.
- `buildRetrySpawnInput` takes provider and model, and keeps its own cap.

The shared helper rather than two copies, as the issue suggested. `check:dups` stays green either way at this size, but the interesting reason is the next one.

## The cap is the one thing infra-lost must not inherit

This is the bit worth reviewing. `decideInfraLostRetry` already gives its retry the original cap **minus what the first attempt spent** (warren-4af7), and that remaining budget comes from the same frozen frontmatter. So spreading the helper wholesale there would replace a $3 remainder with the original $5 cap and hand a failed run a fresh budget.

The helper therefore returns the three slots separately, infra-lost picks two, and its existing cap line stays. There is a test that pins it:

```
keeps the remaining budget as the cap, not the original's full one (warren-0d80)
  original: maxCostUsd 5, costUsd 2  →  maxCostUsdOverride 3
```

The provider retry has no equivalent, because the failure it answers happens before the agent burns the budget it was given, so it inherits the cap verbatim.

## One thing I measured instead of assuming

I first wrote the helper's test asserting that `readMaxCostUsd` rejects a string cap. It does not:

```
"4"   -> 4
-1    -> null
0     -> null
"abc" -> null
```

It coerces a numeric string on purpose, the `cn --fm` trap its own module doc describes. The test now pins the real behaviour and says why, because a retry that silently dropped a cap the original honoured would be its own bug.

## Test plan

- [x] `biome check .` passes
- [x] `tsc --noEmit` passes
- [x] `bun test src/runs/retry`: 40 tests across 3 files, up from 30

The two integration assertions were checked against the old code rather than assumed. With both retry modules stashed and the tests kept:

```
(fail) infra-lost run auto-retry > carries the provider and model the failed run resolved
(fail) createProviderRetryLifecycleExtension > carries the provider, model and cap the failed run resolved
39 pass, 2 fail
```

`check:coverage` is the twelfth gate. It fails on this machine, which is Windows, and fails the same way on unmodified `main`: `comm -3` between the failure sets is empty, 104 distinct tests either way, all POSIX assumptions (`/data/projects` fallbacks, `chmod 0600` round-trips, sqlite file paths, the macOS seatbelt profile).

One run of the full suite also showed `article-ix-gate > a protected path hidden among many ordinary files is still caught` failing. It passes alone (14 pass, 15s: the suite spawns git), and a second full run came back at exactly the baseline 104. Load flake on this machine, not a finding.

## Overlap with #1045

Both touch `provider-retry.ts`. I test-merged rather than guessing:

```
$ git merge --no-ff feat/bounded-provider-retry
Auto-merging src/runs/retry/provider-retry.ts          # clean
Auto-merging src/runs/retry/provider-retry.test.ts
CONFLICT (content): Merge conflict in src/runs/retry/provider-retry.test.ts
```

**The source merges clean.** Only the test file conflicts, because #1045 splits the classifier tests out of it (that file was at 505 lines there) while this one adds tests to it. I deliberately did not repeat that split here, and folded the negative assertion into an existing test instead, so this PR sits at 499 lines and stays small.

Either can merge first. I will rebase whichever loses the race.

I did not exercise a real provider failure against a live cluster.

### DIFF
--- a/src/runs/retry/infra-lost-retry.test.ts
+++ b/src/runs/retry/infra-lost-retry.test.ts
@@ -60,6 +60,8 @@ interface Harness {
 		retryOf?: string;
 		costUsd?: number;
 		maxCostUsd?: number;
+		/** Extra frontmatter frozen onto the run, beside the cap. */
+		frontmatter?: Record<string, unknown>;
 		failureReason?: "sandbox_run_lost" | "crashed" | "provider_error";
 	}) => Promise<RunRow>;
 }
@@ -84,7 +86,10 @@ async function setup(): Promise<Harness> {
 				projectId: project.id,
 				prompt: over.prompt ?? "fix the thing",
 				renderedAgentJson: {
-					frontmatter: over.maxCostUsd !== undefined ? { maxCostUsd: over.maxCostUsd } : {},
+					frontmatter: {
+						...(over.maxCostUsd !== undefined ? { maxCostUsd: over.maxCostUsd } : {}),
+						...(over.frontmatter ?? {}),
+					},
 					sections: {},
 				},
 				trigger: "manual",
@@ -179,6 +184,30 @@ describe("infra-lost run auto-retry (warren-4af7)", () => {
 		const retryEvents = await h.repos.events.listByRun(retry.id);
 		const backlink = retryEvents.find((e) => e.kind === RUN_RETRY_OF_KIND);
 		expect(backlink?.payloadJson).toMatchObject({ retryOf: original.id });
+	});
+
+	test("carries the provider and model the failed run resolved (warren-0d80)", async () => {
+		const original = await h.run({
+			frontmatter: { provider: "openrouter", model: "moonshotai/kimi-k3" },
+		});
+		await hookFor(h)(original.id);
+		expect(h.spawnCalls[0]?.providerOverride).toBe("openrouter");
+		expect(h.spawnCalls[0]?.modelOverride).toBe("moonshotai/kimi-k3");
+	});
+
+	test("keeps the remaining budget as the cap, not the original's full one (warren-0d80)", async () => {
+		// The frozen frontmatter carries the FULL $5 cap. warren-4af7 says the
+		// retry gets what is left, so the inherited cap must not win here.
+		const original = await h.run({ maxCostUsd: 5, costUsd: 2 });
+		await hookFor(h)(original.id);
+		expect(h.spawnCalls[0]?.maxCostUsdOverride).toBe(3);
+	});
+
+	test("passes no provider or model for a run that declared none (warren-0d80)", async () => {
+		const original = await h.run({});
+		await hookFor(h)(original.id);
+		expect(h.spawnCalls[0]?.providerOverride).toBeUndefined();
+		expect(h.spawnCalls[0]?.modelOverride).toBeUndefined();
 	});
 
 	test("uncapped original → retry dispatches with no cap override", async () => {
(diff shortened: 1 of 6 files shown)


jayminwest (approved):
Reviewed after updating onto current main: provider/model/cap inheritance is defensive, infra-lost budget semantics are preserved, and fresh CI is green.

Issue #1173, opened 2026-08-28
Azure DevOps Boards tracker extension (warren-tracker/v1)


PatrickKalkman:
Companion to the ADO forge proposal (#1172). Needed for the deployment whose board is ADO Boards; the other deployment uses Jira and is covered by `extensions/tracker-jira/`. Per the 2026-08-04 decision, trackers after Seeds arrive as external containers behind `RemoteTracker`, so this is a heads-up rather than a request for core changes.

### Plan

- `extensions/tracker-ado/`, modeled on `extensions/tracker-jira/` (#1047): own package, lockfile, tests, image; holds its own ADO PAT; warren never stores it.
- Speaks `warren-tracker/v1`; proved by `extensions/tracker-conformance/`, not by grep.
- Capabilities: `supportsPlans: false`, `isGitNative: false`. Work items map by id (`96379`), state moves on close to a configurable ADO state (`Resolved` / `Closed` per process template), description from the work item plus acceptance criteria.
- The extension only lets a run read a work item and close it on merge. Deciding *when* a work item becomes a run (a board tag as the handoff signal, decline-with-comment when out of scope, PR/stuck comments back on the item) stays in an external controller I run, per the campaign-controller record: it calls `POST /runs` and holds the ADO write credential; core polls nothing.
- WIQL-backed `list` filtered by area path / iteration / assigned-to so the "ready" view is scoped to what the deployment is allowed to pick up.

### Questions

1. Any appetite for it living in-tree beside `tracker-jira`, or would you prefer a third-party repo now that the protocol has two implementations? Either is fine by me; in-tree gets the conformance suite in CI for free.
2. Is the wire contract still "experimental until a foreign implementation survives it unchanged"? If so, I will report every place the ADO mapping needed a protocol change in a FRICTION.md, same as audit-log did.

PR #1180, opened 2026-08-29
tracker: warren-tracker/v1 requires open/closed/other; the bridge rejects anything else

PatrickKalkman:
## Summary

- **The raw-status promise in warren-tracker/v1 does not hold, and it makes every remote tracker unable to finish work.** `src/tracker/remote/protocol.ts` and `docs/design/issue-tracker.md` tell a server to send its raw status string and promise the bridge normalizes it. The bridge (`normalizeIssueStatus`, `src/core/wire-tracker.ts`) only recognizes the literal strings `open` and `closed` and folds everything else to `other`. A remote tracker's `Done` or `Closed` therefore never reads as closed: `src/plan-runs/create.ts` never sees a plan as finished, `src/plan-runs/coordinator.ts:241` never skips a finished child on resume, and `src/runs/reap/auto-plan-run.ts:145` never fires. `extensions/tracker-jira` sends the raw Jira status name, so it is broken this way today; the conformance suite accepts any non-empty string, so it never noticed.
- **Fix: the server folds, the bridge rejects.** Only the server knows whether its `Done`, `Resolved` or `Removed` is terminal, so the protocol now requires every issue `status` to be one of `open | closed | other`. `RemoteTracker` rejects any other string as a malformed payload (`TrackerError: remote tracker returned unknown issue status "Done" for issue X …`) instead of silently folding it — a non-conformant server fails loud on its first read rather than never finishing.
- **The conformance suite now enforces it**, and adds the semantic check that was missing: after `POST /issues/{id}/close`, the issue must read `closed` on both views (`issues/close-reads-closed`). A new self-test proves the suite fails a server that reports raw statuses.
- **`tracker-jira` maps `statusCategory.key`** — `new` → `open`, `indeterminate` → `other`, `done` → `closed`, no category → `other` — the same field it already used to decide whether an issue is terminal.

## Changes

- `src/core/wire-tracker.ts` — `normalizeIssueStatus` doc: for in-core trackers only, and why the bridge must not use it.
- `src/tracker/remote/protocol.ts` — status fields typed `IssueStatus`; header text corrected.
- `src/tracker/remote/remote-tracker.ts` — `requireIssueStatus` replaces the fold; three call sites (issue read, status map, scheduled issues).
- `src/tracker/remote/remote-tracker.test.ts` — rejection tests for `getIssue` and `listIssueStatuses`; fixtures moved onto the vocabulary.
- `src/tracker/conformance-parity.test.ts` — header wording.
- `extensions/tracker-conformance/` — `ISSUE_STATUSES`/`isIssueStatus` in the published protocol copy; vocabulary checks on the status map, the issue read, and scheduled issues; the `issues/close-reads-closed` case; FakeTracker fixture typed; README; self-test for a raw-status server.
- `extensions/tracker-jira/` — `issueStatus()` in `jira/map.ts`, used by the issue read and the search-page status map; `JIRA_NEW_CATEGORY`; protocol copy; README fold table; tests.
- `docs/design/issue-tracker.md` §1 and `CHANGELOG.md` `[Unreleased]`.

Not in this PR: an optional `rawStatus` pass-through for display. Easy to add if you want the tracker's own spelling visible in the UI; left out to keep this to the correctness fix.

## Test plan

- [x] `bun run check:all` — 12/12 with #1179's one-line test fix applied locally; without it, on this macOS host, `check:coverage` fails only on the three `probeSandboxGit` tests #1179 fixes (Linux CI is unaffected)
- [x] `bun test src/tracker src/core` — 144 pass
- [x] `extensions/tracker-conformance`: `typecheck` + `bun test` — 16 pass, including the new raw-status failure case
- [x] `extensions/tracker-jira`: `typecheck` + `bun test` — 67 pass
- [x] The Azure DevOps tracker for #1173 already speaks the vocabulary and passes this suite unchanged

## Compatibility

This is a protocol tightening. A third-party warren-tracker/v1 server that sends raw statuses was already non-functional (nothing it reported could ever be closed); after this change it fails on the first read with a message naming the fix. Both shipped servers (`tracker-jira`, and the ADO tracker for #1173) send the vocabulary.

Relates to #1173.

https://claude.ai/code/session_01Rf5tbgdUDmHzKnNXpiJLvg


luantaraschi (approved):
Approving. Beyond what the description covers, this repairs something that has been broken since the Jira extension merged, and I would rather have that on the record than let it look like a tidy-up.

On `main` the extension puts the raw Jira status name on the wire:

```
extensions/tracker-jira/src/jira/map.ts:101   return issue.fields?.status?.name ?? ""
extensions/tracker-jira/src/jira/map.ts:114   status: statusName(issue)
```

and `normalizeIssueStatus` at `src/core/wire-tracker.ts:42` only recognises the literals `open` and `closed`, so `Done` folds to `other`. Every place that decides closedness reads `status === "closed"`: `coordinator.ts:241`, `create.ts:174` and `:212`, `auto-plan-run.ts:145`. None of them can fire for a Jira-backed issue. What makes it worse is that `isDone` at `map.ts:95` already had the right test, it just was not the one reaching the wire. That is mine, from #1047.

Reading `statusCategory.key` is the right shape too. Jira assigns a category to every status whatever the workflow calls it, so a custom workflow or a non-English site still lands in the vocabulary.

Two small things, neither blocking.

In `src/tracker/remote/protocol.ts` the new import landed between the JSDoc and the constant it belongs to, so the comment now documents the import and `TRACKER_PROTOCOL_VERSION` has none:

```ts
/** The single protocol version warren's bridge speaks. */
import type { IssueStatus } from "../../core/wire.ts";

export const TRACKER_PROTOCOL_VERSION = "warren-tracker/v1";
```

The other is a question. The version string stays `warren-tracker/v1` and `connect()` compares only that string, so an extension image built from `main` today still passes the handshake and then throws on the first read. The header of that file says a mismatch should fail loud at connect rather than per call. Is the intent that tracker images get rebuilt in lockstep? If so it is worth a line in the CHANGELOG entry, which describes the fix but never says rebuild.

PR #1183, opened 2026-08-29
feat: register opencode-go in the provider env registry

dansigma:
## Summary
- Register Pi's `opencode-go` provider in `PROVIDER_ENV_REGISTRY` so `OPENCODE_API_KEY` is forwarded into the sandbox (local/docker) and mapped to an optional `warren-opencode-go-key` Secret on K8s.
- Pi already knows this provider id. Without the registry row, dispatch accepts `provider: opencode-go` and the run dies at auth because the key never enters the allowlist — the same class of miss as warren-81e0 for OpenRouter.

## Test plan
- [x] `bun test src/core/providers.test.ts src/runtime/local/profile.test.ts`
- [ ] Dispatch a Pi run with `provider: opencode-go` and `OPENCODE_API_KEY` set; confirm the sandbox sees the key and the agent authenticates.

Made with [Cursor](https://cursor.com)


luantaraschi (commented):
Entry itself looks right, and the id and env key match pi 0.84.2's own provider docs.

This is the first registry key with a hyphen in it, though, and `resolveProviderSecrets` builds the override variable by interpolating the id uppercased:

```
$ bash -c 'export WARREN_K8S_OPENCODE-GO_SECRET_NAME=x'
export: `WARREN_K8S_OPENCODE-GO_SECRET_NAME=x': not a valid identifier
$ bash -c 'export WARREN_K8S_OPENROUTER_SECRET_NAME=x'
ok
```

The default path is fine, since `warren-opencode-go-key` is a valid Secret name, so what the PR claims does work. What quietly goes missing is the per-provider override that the function's own doc comment describes and that has a test for groq.

I would fix it at the derivation rather than at the entry, something like `provider.toUpperCase().replaceAll("-", "_")`, because pi's list has plenty more of these (`zai-coding-cn`, `kimi-coding`, `qwen-token-plan`) and the next one falls in the same hole.

Separately, pi ships `opencode` on the same `OPENCODE_API_KEY`, and the comment you added to `.env.example` mentions Zen by name. Was leaving that one unregistered deliberate?

PR #1243, opened 2026-09-04
test: report missing kubectl checks as skipped

### LINKED ISSUE #1233: kustomize manifest test passes silently when kubectl is missing
## The defect

The kustomize manifest test passes silently when `kubectl` is missing, so a broken manifest can ship from any environment without the binary, including warren's own agent sandbox image.

`scripts/postgres-component-manifest.test.ts:25-35` probes for `kubectl` with `execFileSync`. When the probe fails, the two render tests print `console.warn("kubectl not found — skipping kustomize render check")` and `return` (`:70-73` and `:101-104`). Bun counts a test that returns as passed, so the summary line and the junit output both report success. Nothing in the run's PR body surfaces the warning either.

`deploy/docker/Dockerfile.agent` installs no `kubectl` (the apt block at `:49` and the node install at `:61-62` are the only package installs), so every warren-dispatched run that touches these manifests gets a green test it never ran.

## Fix direction

Replace the warn-and-return with a real skip so the runner reports it. `test.skipIf(!hasKubectl())(...)` is the pattern already used at `src/db/migrations/drop-placement-tables.test.ts:126`, and `describe.skipIf(...)` at `src/cli/commands/db.test.ts:70`. Compute `hasKubectl()` once at module load and wrap both tests. Bun then prints the skip count in its summary and marks the cases skipped in junit, which is the loud signal this issue asks for.

## Scope

Out of scope: adding `kubectl` to the agent image. That is a one-line change to `Dockerfile.agent` but it is only observable after an image rebuild and deploy, which a contributor cannot verify locally. File it as a follow-up if you want it.

Tests: `scripts/postgres-component-manifest.test.ts` itself. Run it once with `kubectl` on `PATH` and once with it removed from `PATH`, and confirm the second run reports skips rather than passes.

## Getting started

`AGENTS.md` covers setup and conventions. Run `bun run check:all` before pushing. Warnings count as failures.

Tracked internally as `warren-b358`.

joerosbury:
## Summary

- detect `kubectl` availability once when the manifest test module loads
- register the two kubectl-dependent checks with `test.skipIf(...)`
- preserve the existing manifest assertions when `kubectl` is available

Previously, when `kubectl` was unavailable, the two tests printed a warning
and returned normally. Bun and JUnit therefore reported those checks as
passing instead of skipped.

## Validation

Without `kubectl`:

- 1 passed
- 2 skipped
- 0 failed
- JUnit reports `skipped="2"`

With checksum-verified `kubectl` v1.37.0 available:

- 3 passed
- 0 failed
- all manifest assertions executed

Repository validation:

- Warren pre-commit gates: 12/12 passed
- `git diff --check`: passed

The two optional Linux+bwrap sandbox tests were skipped in the controlled
validation environment because `bwrap` was intentionally unavailable; no
bwrap-related source changes are included in this PR.

Fixes #1233

### DIFF
--- a/scripts/postgres-component-manifest.test.ts
+++ b/scripts/postgres-component-manifest.test.ts
@@ -34,6 +34,8 @@ function hasKubectl(): boolean {
 	}
 }
 
+const HAS_KUBECTL = hasKubectl();
+
 interface K8sDoc {
 	kind?: string;
 	metadata?: { name?: string; namespace?: string };
@@ -64,13 +66,9 @@ function renderComponent(): K8sDoc[] {
 }
 
 describe("deploy/k8s/components/postgres", () => {
-	test(
+	test.skipIf(!HAS_KUBECTL)(
 		"kubectl kustomize of an including overlay renders all seven objects",
 		() => {
-			if (!hasKubectl()) {
-				console.warn("kubectl not found — skipping kustomize render check");
-				return;
-			}
 			const docs = renderComponent();
 			const byName = (kind: string, name: string) =>
 				docs.some((d) => d.kind === kind && d.metadata?.name === name);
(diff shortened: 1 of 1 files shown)


jayminwest:
Great work @joerosbury, thanks for the PR!

PR #1249, opened 2026-09-06
fix(server): reject POST /runs with an unknown seedId (#1234)

### LINKED ISSUE #1234: POST /runs accepts a seedId the tracker does not know and dispatches anyway
## The defect

`POST /runs` accepts a `seedId` that does not exist in the project's issue tracker and dispatches anyway. The agent then works from a dangling id, and when it runs the seeds CLI to update "its" issue it creates a duplicate instead. Nothing on the request path fails.

`src/server/handlers/runs/dispatch.ts:143` reads `seedId` from the body and forwards it unchanged at `:195`. `src/runs/spawn/dispatch.ts` freezes it onto the run row with no lookup. The only tracker call is the post-spawn metadata write in `src/runs/spawn/seed-extensions.ts:110-117`, which swallows any error into a `seeds_extension_write_failed` event. So a missing seed dispatches successfully and the caller gets a 201.

The tracker seam already has the lookup: `IssueTracker.getIssue` (`src/tracker/contract.ts:70`) throws `IssueNotFoundError` (`src/core/wire-tracker.ts:159`) for an unknown id.

## Fix direction

Validate before any side effect. `src/plan-runs/create.ts:171` and the comment at `:187-190` show the shape: every posted issue id goes through `tracker.getIssue`, and a missing one throws `IssueNotFoundError`, which the plan-run handler renders as 404. Do the same in the run dispatch handler when `seedId` is present and `deps.issueTracker` is available, before `spawnRun` is called.

`src/server/errors.ts:218-232` has no arm for `IssueNotFoundError` today, so it would render as 500. Add `if (err instanceof IssueNotFoundError) return 404;` beside the other 404 mappings. Match the plan-run precedent and use 404 rather than 400.

## Scope

Out of scope: validating seed ids on the scheduler or trigger dispatch paths, and any change to what the agent does with the id once it has it.

Tests: `src/server/handlers/runs.dispatch.test.ts` already has an `optional seedId persists` case at `:192`; add a sibling that posts an unknown id and expects 404 with no run row created. Add a case to `src/runs/spawn/seed-extensions.test.ts` only if you touch that file. If the error envelope shape changes, the golden at `src/server/__golden__/responses` is the pin.

## Getting started

`AGENTS.md` covers setup and conventions. `bun test src/server/handlers/runs.dispatch.test.ts` runs the handler suite in-process with no credentials. Run `bun run check:all` before pushing. Warnings count as failures.

Tracked internally as `warren-8fd5`.

yegorske50:
Closes #1234

## Summary

- Validates a dispatched `seedId` against the wired `IssueTracker` before `POST /runs` has any side effects, instead of only discovering it's missing in the post-dispatch metadata write (whose failures are silently swallowed today).
- Maps the tracker's `IssueNotFoundError` to a 404 `issue_not_found` response.
- No tracker wired → validation is skipped, matching the existing `resolveSeedTracker` "neither wired → no write" precedent, so a project without a tracker configured is unaffected.

## Changes

- `src/server/handlers/runs/dispatch.ts` — new `validateSeedId` helper, called in `buildHttpSpawnOptions` right after `projectId`/`seedId` are resolved and before `spawnRun` fires.
- `src/server/errors.ts` — `IssueNotFoundError` now maps to 404 in `warrenStatusFor` (it previously fell through to the generic 500 despite being a `WarrenError` subclass, since it wasn't in the status table).
- `src/server/handlers/runs.dispatch.seed-validation.test.ts` — new test file, following the existing `runs.dispatch-idempotency.test.ts` / `runs.dispatch.base-commit.test.ts` split rather than growing `runs.dispatch.test.ts` past the repo's 500-line file-size budget.

## Test plan

- [x] `bun run lint` passes
- [x] `bun run typecheck` passes
- [x] `bun test src/server/handlers/runs.dispatch.test.ts src/server/handlers/runs.dispatch.seed-validation.test.ts` passes (12/12)
- [x] Reproduced the bug first: reverted the fix locally, confirmed the new test fails against unmodified `main` (`Expected: 404, Received: 201`), then restored the fix and confirmed green.

- [ ] `bun run check:all` — fails locally only on `check:coverage`'s `src/sandbox/proxy-server.test.ts` (needs real control over HTTP forwarding on loopback, which my local sandboxed dev environment's own outbound proxy interferes with); reproduces identically on a clean `main` checkout with zero changes, so it should pass under CI's real network.


luantaraschi (approved):
Approving and merging this. Jaymin is away until September 14 and asked me to merge linked PRs that pass CI, so this does not need to wait for him.

The fix matches what #1234 asked for, and the test pins the half that matters: not just the 404, but that no run row and no sandbox call happened.

I ran a differential against `main` before merging, with a stub `IssueTracker` swapped in for the seeds one, to see everything the new `getIssue` call changes:

| `getIssue` behavior | `main` | this PR |
| --- | --- | --- |
| throws `IssueNotFoundError` | 201, run created | 404 `issue_not_found`, nothing created |
| resolves the issue | 201, run created | 201, run created |
| throws a plain `TrackerError` | 201, run created | 500 `tracker_error` |

The third row is new behavior that #1234 did not ask for, and I checked the house posture before reading it as a problem. `src/server/handlers/projects.ts:211` already documents "TrackerError from `getIssue` bubbles up as 500" for the single-seed read, so this is consistent with the repo rather than a departure from it. Not a blocker.

Two notes for a follow-up, or for whenever someone next opens the file. Neither changes my read of the fix.

The `deps.issueTracker === undefined` early return never fires on a real boot. `src/server/main/deps.ts:203` is `issueTracker: issueTracker ?? new SeedsTracker({ sdBinary, spawn: defaultSpawn })`, so the field is always populated outside tests that assemble deps by hand. That makes the comment above the helper the part I would change: it says an untracked project's dispatch stays unaffected, and an untracked project is exactly the case that now reaches the tracker and can come back 500. The code is fine, the promise in the comment is not.

The second is `requireTrackerProject` at `src/server/handlers/projects.ts:161`. Its own docstring calls it the shared three-gate seeds-read preamble that every seeds-read handler starts from, and `validateSeedId` is a new seeds read that does not use it. Going through it would drop the hand-rolled `projects.require` and would turn a project with no `.seeds/` into a 400 `ProjectLacksTrackerError` instead of a 500. That is a wider behavior change than the issue asked for, which is why I am not asking for it in this PR. Worth raising with Jaymin when he is back.

On your `check:all` note: CI ran all four checks green on `a8a6e431`, which is the head this merges, so `src/sandbox/proxy-server.test.ts` looks like your local network rather than the change. Thanks for reverting the fix and confirming the test failed first. That is the part that made this quick to trust.

PR #1256, opened 2026-09-10
fix(workspace): keep harness state and seed drops out of run branches

### LINKED ISSUE #1239: run branches commit .pi/sessions and .warren/agent.json when the target repo's .gitignore does not cover them
## The defect

When the target repository's `.gitignore` does not cover warren's harness files, the agent's own commits sweep those files into the run branch, and they land in the upstream pull request.

The pi harness writes its session transcript inside the git worktree. `src/runtime/adapters/pi-argv.ts:58` pins `PI_SESSION_DIR = ".pi/sessions"` and the argv at `:87-88` passes it as `--session-dir`, relative to the workspace. `src/runtime/adapters/pi.ts:133` creates that directory under `ctx.workspacePath`. Warren's seed also writes `.warren/agent.json` into the worktree, and `src/runs/seed.ts:38` assumes it is gitignored, which is true for warren's own repo and false for a mirror of someone else's.

Nothing prevents a commit from including them:

- `pi.ts:66` declares `harnessStatePrefixes: [".pi/sessions/"]`, but the only consumer is `src/runs/reap/util.ts:183`, which uses the list to classify a dropped commit after the fact.
- The seed-reset step (`src/runtime/local/finalize-seed-reset.ts`, `src/runs/reap/seed-reset.ts:43-55`) reverts only files warren seeded and only when their bytes still equal the seed. `.pi/sessions/*.jsonl` is never seeded, so it is never reset. The K8s path has no seed-reset step at all.
- No code writes `.git/info/exclude` (zero hits in `src/`).

Observed on 2026-08-31: a run against an external mirror committed `.pi/sessions/*.jsonl` and `.warren/agent.json` into the PR branch.

## Fix direction

Seed `.git/info/exclude` when the worktree is materialized. `src/workspace/materialize.ts` is where `git worktree add` runs; after it, write one line per prefix from `harnessStatePrefixes()` (`src/runtime/adapters/index.ts:76`) plus the seed paths that `buildSeedFiles` in `src/runs/seed.ts:74` produces. For a `git worktree` the exclude file lives under the parent repo's `.git/worktrees/<name>/info/exclude`; use `git rev-parse --git-path info/exclude` to find it rather than composing the path. Excludes are local to the checkout and never reach the remote, which is the property this needs.

Do this for both the local and the K8s workspace paths, since the K8s path clones inside the pod (`src/runtime/k8s/workspace-init.ts`).

## Scope

Out of scope: moving pi's session directory out of the worktree entirely. That would also work, but `pi.ts:139` reads the newest session file from the workspace to recover the session id, and the read would have to move too. A follow-up can do it.

Tests: `src/workspace/materialize.test.ts` (assert the exclude file contents after materialization against a temp repo), `src/runtime/adapters/pi.test.ts` if the prefix list changes, and a K8s workspace-init unit test for the same write.

## Getting started

`AGENTS.md` covers setup and conventions. The workspace tests run against temporary git repositories with no network. Run `bun run check:all` before pushing. Warnings count as failures.

Tracked internally as `warren-194a`.

AlsoTheZv3n:
Closes #1239

## What

Seed `info/exclude` when a run's workspace is materialized, so the harness's own scratch and warren's seed drops stay out of the agent's commits. A target repository whose `.gitignore` does not cover those paths no longer carries `.pi/sessions/*.jsonl` or `.warren/agent.json` into the pull request.

Covered on both paths the issue names: the LocalProvider engine and the K8s init container.

## One correction to the fix direction

The issue says the exclude file for a worktree lives at `.git/worktrees/<name>/info/exclude`. git does not read that file. `info/exclude` is a common path, so it stays shared across every worktree of a clone. The recommended command is still the right one: `git rev-parse --git-path info/exclude` returns that shared file.

Verified on git 2.43.0, with `.pi/sessions/s.jsonl` and `.warren/agent.json` present in a worktree, writing the patterns to one file at a time:

| file written | `git status --porcelain` in the worktree |
| --- | --- |
| `.git/worktrees/wt/info/exclude` | `?? .pi/` and `?? .warren/` |
| `.git/info/exclude` | empty |

`src/workspace/git-exclude.test.ts` pins both halves of that: one test asserts the worktree goes clean, and asserts the bytes landed in the parent clone's `.git/info/exclude` rather than under `.git/worktrees/`.

Full detail and the reproduction are in the issue thread.

## Shape

Because the write reaches the host clone, it is confined to a marked block and rendered in full on every materialization:

```
# >>> warren managed >>>
.claude/
.claude.json
.pi/sessions/
.warren/agent.json
# <<< warren managed <<<
```

Every run renders the same bytes, so two runs against one project converge instead of racing, and an operator can see and delete exactly what warren added. Operator lines outside the block survive untouched, and an empty pattern list removes the block rather than leaving a stale one. Excludes are local to a checkout and never reach the remote, which is the property this needs.

The clone-backed paths (`materializeViaClone`, and the pod's per-run clone) each own their `.git`, so neither shares nor outlives anything.

## Where the patterns come from

The callers supply them, rather than `src/workspace/` reaching into the runtime:

```ts
await writeWorkspaceExcludes(workspacePath, [
	...harnessStatePrefixes(),
	...spec.seedFiles.map((file) => file.path),
]);
```

`src/workspace/materialize.ts` documents itself as neutral and deliberately free of coupling, so importing `src/runtime/adapters/` and `src/runs/seed.ts` into it would cut against that. `writeWorkspaceExcludes` takes optional `git` and `fs` seams shaped to match the init container's existing `InitFs` and `InitGitRunner`, so the pod path hands its injectable fakes straight in and stays testable without disk or a real git.

`writeSeedFiles` now returns the paths it wrote, which is what the K8s side passes as the seed half of the list.

## Tests

- `src/workspace/git-exclude.test.ts`, new: 13 cases. Seven cover the block algebra as a pure function (append, replace rather than stack, idempotence, operator lines before and after, dedupe, removal). Six run against real temporary git repositories: a plain clone, a worktree, an operator's pre-existing entries, a repeat seeding, no temp file left behind, and a loud failure when the workspace is not a repository.
- `src/runtime/k8s/workspace-init.test.ts`: a new case asserting both halves of the list reach the pod's exclude file.
- `src/runtime/local/engine.test.ts`: the existing create-path test now asserts the seeded `.warren/agent.json` does not show up in `git status`.

Two test-harness changes came with it. The fake git runners now answer `rev-parse --git-path`, which real git always answers and the blanket success stub did not. And five `runWorkspaceInit` cases now pass the no-op `fs` explicitly; they were only staying off disk because the seed loop returns early without a manifest, so the real `fs` default was never exercised.

`bun run check:all` passes 12 of 12 gates, measured on Ubuntu 24.04 with bubblewrap 0.9.0.

One note on reproducing that: on a machine with bwrap, `check:coverage` also fails on `src/sandbox/git-preflight.test.ts`, which this branch does not touch and which fails identically on an unmodified `main`. That defect is #1255, sent separately. This branch does not depend on it; the two were measured together only because the local pre-commit gate runs the whole manifest.

## Out of scope

Moving pi's session directory out of the worktree, as the issue notes.

Two things noticed and deliberately left alone, each worth its own change: the header comment on `src/workspace/materialize.ts` still says the module is "intentionally unimported by the domain", which `src/runtime/local/engine.ts` has since made untrue; and `scripts/bundle-size-budgets.json` repeatedly recommends `bun run check:bundle-size:build --update`, a script that is not in `package.json`. Happy to file either.

### DIFF
--- a/src/runtime/k8s/workspace-init.test.ts
+++ b/src/runtime/k8s/workspace-init.test.ts
@@ -11,6 +11,17 @@ import {
 const ok = { exitCode: 0, stdout: "", stderr: "" };
 
 /**
+ * Real git answers `rev-parse --git-path <p>` with a path, and the exclude
+ * seeding (warren-194a) reads it. A fake that returned the blanket `ok` would
+ * hand back an empty path, which is a failure real git never produces.
+ */
+function gitPathReply(args: string[]): { exitCode: number; stdout: string; stderr: string } | null {
+	const i = args.indexOf("--git-path");
+	if (args[0] !== "rev-parse" || i === -1) return null;
+	return { exitCode: 0, stdout: `.git/${args[i + 1] ?? ""}\n`, stderr: "" };
+}
+
+/**
  * Recording git that lets a test steer the mirror probe + fail specific argv.
  * `bareRepoProbe` decides what `git rev-parse --is-bare-repository` reports
  * (default `false` ⇒ mirror treated as absent ⇒ create path).
@@ -25,16 +36,24 @@ function cacheGit(overrides: {
 		if (overrides.fail?.(args, opts?.cwd)) {
 			return Promise.resolve({ exitCode: 1, stdout: "", stderr: "boom" });
 		}
-		if (args.includes("--is-bare-repository")) {
-			return Promise.resolve({
-				exitCode: overrides.bareRepoProbe ? 0 : 128,
-				stdout: overrides.bareRepoProbe ? "true\n" : "",
-				stderr: overrides.bareRepoProbe ? "" : "fatal: not a git repository",
-			});
-		}
-		return Promise.resolve(ok);
+		return Promise.resolve(
+			gitPathReply(args) ?? bareRepoReply(args, overrides.bareRepoProbe === true) ?? ok,
+		);
 	};
 	return { git, calls };
+}
+
+/** What `git rev-parse --is-bare-repository` reports for the steered mirror. */
+function bareRepoReply(
+	args: string[],
+	isBare: boolean,
+): { exitCode: number; stdout: string; stderr: string } | null {
+	if (!args.includes("--is-bare-repository")) return null;
+	return {
+		exitCode: isBare ? 0 : 128,
+		stdout: isBare ? "true\n" : "",
+		stderr: isBare ? "" : "fatal: not a git repository",
+	};
 }
 
 const CACHE_ENV = {
@@ -116,6 +135,8 @@ function recordingGit(overrides: { fail?: (args: string[]) => boolean } = {}): {
 	const git: InitGitRunner = (args, opts) => {
 		calls.push({ args, ...(opts?.cwd !== undefined ? { cwd: opts.cwd } : {}) });
 		if (overrides.fail?.(args)) return Promise.resolve({ exitCode: 1, stdout: "", stderr: "boom" });
+		const gitPath = gitPathReply(args);
+		if (gitPath !== null) return Promise.resolve(gitPath);
 		return Promise.resolve(ok);
 	};
 	return { git, calls };
@@ -132,7 +153,7 @@ describe("runWorkspaceInit", () => {
 				WARREN_WORKSPACE_PATH: "/ws",
 				WARREN_GIT_TOKEN: "tok",
 			},
-			{ git, log: () => {} },
+			{ git, fs: noopFs, log: () => {} },
 		);
 		expect(calls[0]?.args).toEqual([
 			"clone",
@@ -158,9 +179,10 @@ describe("runWorkspaceInit", () => {
 				WARREN_BRANCH: "b",
 				WARREN_BASE_BRANCH: "main",
 			},
-			{ git, log: () => {} },
+			{ git, fs: noopFs, log: () => {} },
 		);
-		expect(calls.map((c) => c.args[0])).toEqual(["clone", "switch"]);
+		// The trailing rev-parse is the exclude seeding resolving info/exclude.
+		expect(calls.map((c) => c.args[0])).toEqual(["clone", "switch", "rev-parse"]);
 	});
 
 	test("branch === baseBranch (ref-dispatch, warren-dac8) skips the colliding switch -c", async () => {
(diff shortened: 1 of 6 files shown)


AlsoTheZv3n:
Closing this. I verified it properly after opening it, and the approach is wrong, not just the details. Nothing here should reach a review queue.

Five defects, each reproduced:

**1. `.claude/` is a directory exclude, and it swallows real work.** The pattern list is `harnessStatePrefixes()`, which is `[".claude/", ".claude.json", ".pi/sessions/"]`. That list exists for reap's dirty-path classifier (`src/runs/reap/util.ts:183`), where a broad prefix is harmless. As a git exclude it is not. This repo tracks five files under `.claude/`, and any *new* file there disappears:

```
$ printf '# >>> warren managed >>>\n.claude/\n...' > .git/info/exclude
$ echo neu > .claude/commands/brandnew.md && echo arbeit > agent-work.txt
$ git status --porcelain
?? agent-work.txt
$ git add -A && git diff --cached --name-only
agent-work.txt
```

Exit 0, no warning, the file is gone. The pi adapter avoids exactly this by pinning `.pi/sessions/` rather than `.pi/`, and `src/runtime/adapters/index.test.ts:64` asserts `not.toContain(".pi/")`. I did not carry that narrowing over.

**2. The local path writes the host clone's shared file, so runs overwrite each other.** `renderManagedExclude` replaces the block wholesale. The pattern list is per-run, because the seed half comes from `buildSeedFiles(agent)`. Two runs with different agents render different bytes, so run B drops run A's seed exclusions while A is still going, and A's seed drops become visible again. That is the pollution this issue is about, reintroduced non-deterministically.

**3. Concurrent runs in one server process kill the run.** The temp path is `${excludePath}.warren-${process.pid}.tmp`. The local engine dispatches runs inside one process, so the pid is identical, the two writers collide, and the second `rename` throws ENOENT out of `LocalEngine.create`.

**4. The local path has no effective test.** My assertion was `expect(dirty.stdout).not.toContain(".warren/agent.json")`. `git status --porcelain` collapses an untracked directory to `?? .warren/`, so that assertion passes with the fix removed. Article IV.

**5. The PR body's "12 of 12 gates" is wrong.** I measured before rebasing the commit onto `main`. On the branch as pushed it is 11 of 12.

Also: the leftover block makes `applyCloneDeltas`' `git add -- <pathspec>` exit 1 (`src/runs/reap/clone-apply.ts:102`); `git clean -f` in `finalize-seed-reset.ts` becomes a no-op for the newly excluded seed drops; a half-written block (START without END) deletes operator lines below it; and operator lines *after* the block get reordered ahead of it, which changes gitignore precedence.

The shared `info/exclude` cannot carry this. A worktree-scoped `core.excludesFile` can — I measured it working and genuinely scoped earlier in the issue thread. That is a design call for you rather than something to slip into a fix, so the findings are on #1239 for whoever picks it up.

Sorry for the noise.

